PR #3, opened 2015-09-08
Rename files to follow standard as discussed in #1

krolow:
This commit attempt to close issue #2, by following a nomenclature standard:

1 - Unix style for filenames
2 - No suffix for filenames

PR #4, opened 2015-09-10
React router

maxcnunes:
To make work the react-router properly I had to remove the references to material-ui. Because of some problem with the material-ui context. Probably there is a way to make them work together. But since we were already talking in take away the material and replace it with another layout I just did what was easier.

Still have some material-ui code commented and package references because I'm thinking in remove them in another pull request. Which one I'm going to configure another base layout.


krolow:
looks good @maxcnunes you gonna need just to rebase master code once we have merged the files rename of PR #3

PR #5, opened 2015-09-12
creates a basic connection page

maxcnunes:
it is the initial page and clicking in any connection moves to database list page


src/actions/connections.js
```diff
@@ -0,0 +1,29 @@
+import { LODAD_CONNECTIONS } from '../constants/action-types';
+
+
+export function loadConnections() {
+  const connections = [
+    {
```
krolow: perhaps your stub can already follow the one proposed: https://github.com/sqlectron/sqlectron/wiki/Database-connections-file-descriptor

krolow:
looks good, the only thing its about stub that perhaps can follow the one proposed by @esnunes

:+1:

maxcnunes:
Ok.. I will let these changes related to connections to who is going to implement the loading of connections from the configuration file.

PR #6, opened 2015-09-13
Connection container

maxcnunes:
First version of the "main" page.
- it is loading databases trough sqlectron-db
- it is loading tables trough sqlectron-db
- it is executing query trough sqlectron-db

screen shot 2015-09-13 at 6 13 31 pm

A lot of stuff to do yet. But I think is better break the rest in different stories.


krolow:
looks good :+1:

PR #8, opened 2015-09-16
Upgrade all dependencies

maxcnunes:
Principle changes:
- React 0.14.*
- Redux 3.0.*
Removed:
- lodash
- moment
Included object-assign because of a problem with Radium. Remove it in the 
future.


krolow:
:+1:

PR #9, opened 2015-09-16
Configure application menu and shortcuts

maxcnunes:
Also was add a shortcut to execute the query by shortcut.


krolow:
are commands of windows/linux cover in this commit or should take a issue a part of that to handle thoses?

maxcnunes:
Since I use `CmdOrCtrl` key it should work in any platform or keyboard. Could you find any command that wouldn't work in some platform?

PR #10, opened 2015-09-17
Rename actions types, actions and reducers

maxcnunes:
Applies a common patter name for those items


src/pages/database.jsx
```diff
@@ -20,36 +20,36 @@ export default class DatabaseList extends ValidatedComponent {
...
   onDiscQueryClick() {
     React.findDOMNode(this.refs.queryBoxTextarea).value = '';
   }
 
-  buildQueryResult(queryResult) {
-    if (queryResult.error) {
-      return <pre>{JSON.stringify(queryResult.error, null, 2)}</pre>;
+  buildQueryResult(query) {
+    if (query.error) {
+      return <pre>{JSON.stringify(query.error, null, 2)}</pre>;
```
krolow: why we have this pre in case of error?

maxcnunes: The current error object that is coming from Knex is not giving much useful information. While we don't improve this error message I though in just display the whole raw error object and using `pre` we keep a better indentation for the output object. But this is just temporary, in the future we should display a label with useful error messages.

PR #11, opened 2015-09-17
Configure code editor for syntax highlighting

maxcnunes:
Is much more fun write SQL code with syntax highlight :)

ace-code-editor


krolow:
:+1:

PR #14, opened 2015-09-20
Replace react-hot-loader

maxcnunes:
The react-hot-loader has been deprecated. The new module is react-transform-hmr.


krolow:
:+1:

PR #15, opened 2015-09-20
Configure databases for dev environment

maxcnunes:
It includes docker-compose to setup dev environment with different databases.


krolow:
:+1:

PR #16, opened 2015-09-20
Add filter box to filter databases and tables

maxcnunes:


krolow:
it wouldn't be better if only update the list instead of have to handle visibility=true/false ?

i mean return a new list without the things that does not fit the search... so we do not need to handle this loadAll and add the if statement to display block/none

maxcnunes:
Initially I implemented such as you said. But I realised once you have changed the state filtering the list you would not have the initial state with all items anymore. This was the only way I find out filtering the list without lose the whole list. Do you have any idea how better implement it?

maxcnunes:
Can I merge it?

krolow:
yes, go for it :+1:

PR #17, opened 2015-09-20
Organize db session

maxcnunes:
Moves db’s main process code to a proper file. And makes actions not aware of main process.

### DIFF
--- a/src/actions/databases.js
+++ b/src/actions/databases.js
@@ -1,4 +1,4 @@
-const getDB = require('remote').require('./src/db').getDB;
+import { getDBSession } from '../db';
 
 
 import {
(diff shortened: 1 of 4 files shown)


src/actions/databases.js
```diff
@@ -1,4 +1,4 @@
-const getDB = require('remote').require('./src/db').getDB;
+import { getDBSession } from '../db';
```
krolow: about the name, "session" looks a bit strange for me... perhaps we can use getDBConnection or some other thing, it's more a database connection or connection instance than a session...

what do you think?

maxcnunes: I didn't like the name either. I had tought in name it `getDBConnection` as well but actually it is more than only a database connection. It is the sqlectron-db that is a wrap over knex.js. In the end I chose Session inspired on [nhinbernate](http://nhibernate.info/doc/nh/en/index.html#quickstart-playingwithcats).

@esnunes what do you think?

esnunes: I think the concept of a `Server Session` fits to the requirements due to the fact that the *back-end* of the application must handle tunnels and connections issues like a drop of a connection.

maxcnunes: So is it gonna be `getServerSession`?

maxcnunes: Merging it to master. We can pick up a better name later.

PR #18, opened 2015-09-21
Create header menu

maxcnunes:
Contains: Logo, current server and db and close connection button

header

PS: The breadcrumb still static. To implement it properly we have to decide in which store we gonna keep the selected server and database. Lets decide it after finish the connection management screen.

### DIFF
--- a/src/components/header.jsx
+++ b/src/components/header.jsx
@@ -1,0 +1,35 @@
+import React, {Component, PropTypes} from 'react';
+import { Link } from 'react-router';
+
+
+export default class Header extends Component {
+
+  render() {
+    return (
+      <div className="ui top fixed menu borderless">
+        <div className="item">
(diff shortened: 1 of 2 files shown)


krolow:
i've also worked in similar feature, i didn't PR over the weekend because i was not happy with the end result... maybe you can check it here: https://github.com/krolow/sqlectron/commit/17df4845bd6d9e98574cf50d758d1cd8e4299731

screenshot from 2015-09-21 08 38 46

maxcnunes:
I visually prefer mine because it is more compact and simpler. But in case you prefer your own solution I'm ok using yours. Also your solution looks like don't show the current connected database, then we would need show this information somewhere.

maxcnunes:
Merging it to master for now. We can change it later.

PR #22, opened 2015-10-20
Manage servers

maxcnunes:
Well... What can I say about this commit? I buch of stuff.
But almost everything was related to the server management screen.
There is some improvements we can do in the future. But I guess we can consider this as completed for now.

splash
all
filter
new
validation
edit
remove


README.md
```diff
@@ -1,4 +1,4 @@
-![SQLECTRON Logo](https://www.dropbox.com/s/038vmsywakz2hug/SQLECTRON-400px.png?raw=1)
+![SQLECTRON Logo](https://www.dropbox.com/s/dt06tif4xo1khbf/SQLECTRON-400px.png?raw=1)
```
krolow: It would't be better to place into the app the image, i can imagine the scenario where we are without internet connection and it might not load the logo image

src/renderer/components/server-list.jsx
```diff
@@ -0,0 +1,59 @@
...
+
+  groupItemsInRows(items) {
+    const itemsPerRow = 4;
+    return items.filter(item => item.visible).reduce((rows, item, index) => {
+      const data = { id: index, server: item };
+
+      const position = Math.floor(index / itemsPerRow);
+      if (rows[position]) {
+        rows[position].push(data);
+      } else {
```
krolow: maybe can avoid else by:

```javascript
if (!rows[position]) {
  rows[position] = [];
}

rows[position].push(data);
```

README.md
maxcnunes: Yes. I am keeping it outside of the app just before the first release. To not increase the size of the git repository in case we change the images.

src/renderer/components/server-list.jsx
maxcnunes: Sure.

PR #92, opened 2016-02-17
Update react-resizable.css

kbradl16:
removes scroll bar
scrollbars

Removes scroll bar at bottom of sidebar ResizableBox.


maxcnunes:
This change is related with your other pull request. I will reject it here and include again after some changes I have to do in your other pull request.

PR #93, opened 2016-02-17
Update semantic.css

kbradl16:
Removes horizontal scroll bar


maxcnunes:
It is a vendor file. We should avoid if possible changing it. I will include this change in another file.

Issue #65, opened 2015-12-09
Use fixed left panel


maxcnunes:
Do not scroll with the right panel.

PR #94, opened 2016-02-18
Reduce amount of scroll bars.

kbradl16:
Make the app feel more like 1 part instead of 2 parts.

Reduce amount of scrolling, reduce annoyance factor. :) 

p.s. Love the app!!!

scrollbar galore
Lots of scrolling

2 scroll bars
After My edits

### DIFF
--- a/src/renderer/containers/query-browser.jsx
+++ b/src/renderer/containers/query-browser.jsx
@@ -24,8 +24,8 @@ require('../components/react-tabs.scss');
 const STYLES = {
   wrapper: { paddingTop: '50px' },
   container: { display: 'flex', padding: '10px 10px 50px 10px' },
-  sidebar: { overflowY: 'auto', height: '100vh' },
-  content: { flex: 1, overflow: 'auto', height: '100vh' },
+  sidebar: { overflowY: 'auto' },
+  content: { flex: 1, overflow: 'auto', height: '100%' },
 };
 
 
--- a/src/renderer/containers/query-browser.jsx
+++ b/src/renderer/containers/query-browser.jsx
@@ -22,7 +22,7 @@ require('../components/react-tabs.scss');
 
 
 const STYLES = {
-  // wrapper: { paddingTop: '50px' },
+  wrapper: {  },
   container: { display: 'flex', height: '100vh', box-sizing: 'border-box', padding: '50px 10px 40px 10px'  },
   sidebar: { overflowY: 'auto' },
   resizeable: { max-width: '100%', padding-right: '5px' },
--- a/src/renderer/containers/query-browser.jsx
+++ b/src/renderer/containers/query-browser.jsx
@@ -22,10 +22,11 @@ require('../components/react-tabs.scss');
 
 
 const STYLES = {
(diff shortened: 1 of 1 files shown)


maxcnunes:
Awesome contributing. Right now I am on vocations. So I will take a look on this when I get back to work. Mean while could you check if the sidebar still scrolling separately from the right? It was a feature some people asked me (https://github.com/sqlectron/sqlectron-gui/issues/65).

Thanks!

kbradl16:
Ah, that makes sense... 
I've come up with a better solution to keep the 2 separate scroll bars and still reduce the overall amount.

kbradl16:
How it looks after edits
newlook

src/renderer/containers/query-browser.jsx
```diff
@@ -22,10 +22,11 @@ require('../components/react-tabs.scss');
 
 
 const STYLES = {
-  wrapper: { paddingTop: '50px' },
-  container: { display: 'flex', padding: '10px 10px 50px 10px' },
-  sidebar: { overflowY: 'auto', height: '100vh' },
-  content: { flex: 1, overflow: 'auto', height: '100vh' },
+  wrapper: {  },
+  container: { display: 'flex', height: '100vh', box-sizing: 'border-box', padding: '50px 10px 40px 10px'  },
```
maxcnunes: `box-sizing` is not a valid react style attribute https://facebook.github.io/react/tips/inline-styles.html

src/renderer/containers/query-browser.jsx
```diff
@@ -22,10 +22,11 @@ require('../components/react-tabs.scss');
...
+  sidebar: { overflowY: 'auto' },
+  resizeable: { max-width: '100%', padding-right: '5px' },
```
maxcnunes: `max-widht` and `paddign-right` are not valid react style attributes

src/renderer/containers/query-browser.jsx
```diff
@@ -22,10 +22,11 @@ require('../components/react-tabs.scss');
...
+  content: { flex: 1, overflow: 'auto', padding-left: '5px' },
```
maxcnunes: `padding-left` is not a valid react style attribute

src/renderer/containers/query-browser.jsx
```diff
@@ -269,7 +270,7 @@ class QueryBrowserContainer extends Component {
             <ResizableBox className="react-resizable react-resizable-ew-resize"
               width={235}
               minConstraints={[235, 300]}
-              maxConstraints={[750, 10000]}>
+              maxConstraints={[750, 10000]} style={STYLES.resizeable}>
```
maxcnunes: `ResizableBox` does not support `style` property. Actually it is crashing the `ResizableBox`.

maxcnunes:
Even with some problems in your pull request I will accept it and apply the fixing myself since I already know how to fix them.

maxcnunes:
Done. Let me know if it still working as you were expecting.

kbradl16:
Ooops, sorry bout that.  I was making my best guesses... I just assumed it was like valid css styles.

Issue #90, opened 2016-02-10
Doesn't show value for boolean columns


brunolm:
I have a boolean column, I know for sure that the value is `true` but it doesn't show me that, instead the column is empty.

PR #99, opened 2016-03-21
Force result in row be always string

krolow:
For boolean make sense to be always string...

**So it will looks like:**

image

**Before it was showing:**

image

### DIFF
--- a/src/renderer/components/query-result-table.jsx
+++ b/src/renderer/components/query-result-table.jsx
@@ -56,7 +56,7 @@ export default class QueryResultTable extends Component {
       return (
         <tr key={index}>
           {columnNames.map(name => {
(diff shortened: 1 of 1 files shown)


esnunes:
Can you check for null values before converting to string? Another option would be using `JSON.stringify(value)` in this case it would be possible to see if the value is `true` or `"true"`

maxcnunes:
It has already been fixed in v1.5.0. https://github.com/sqlectron/sqlectron-gui/commit/b878fbd19e32a31636dd936c8aa4dba387d37d3a :dancers:

krolow:
**errrouuu**, it is already fixed :laughing: at #90 

so closing the issue

Issue #98, opened 2016-03-17
UI: View stored procedures in DB menu


darkwisperer:
Enhancement Request:
Ability to view a list of stored procedures from the left hand  DB menu probably similar styling to the DB name just in its list. Similar to how its done in MSSQL management studio
![MSSQL management studio](http://i.stack.imgur.com/ijmaf.png)
, there would be a collapsed list that would have the results from
`select *  from information_schema.routines  where routine_type = 'PROCEDURE'`

Clicking on the name of the SP would open a tab with the SP's code there.

So it needs to be decided if SPs, Functions, Triggers should be in their own collection to separate them from the tables. 
-Should we combine the SPs, Functions, Triggers in to a programmability (or some other name) folder and leave the structure the way it is?
- should the tables be in a collection with SPs, Functions, Triggers also being separate collections like SQLdeveloper ![SQL Dev](http://www.oracle.com/ocom/groups/public/@otn/documents/digitalasset/118334.gif)

I'm not a fan of having a programmability folder/collection, but it can minimize the clutter that is SQLDeveloper. Another option could be a context menu on the DB name that has the option to see the SPs and click the one you with to view...

It is a lot to go over and think about. With direction I would consider helping with this in a couple months

BornaP:
I'm glad you opened ticket for this issue, because I wanted to do the same :+1: 

I think structure similar to one in mysql workbench is optimal here. I would keep triggers (and I would add columns) collapsed under each table and views, SP and functions should be in their own collection to separate them from the tables. Something like this:
![sql_metadata_structure](https://cloud.githubusercontent.com/assets/8470710/13947155/11151ace-f019-11e5-8ea0-44c08cd1c08e.png)

I'll try to contribute on this in a next couple of days/weeks. Also, @maxcnunes ,could you put 'Enhancement' label to the issue.

maxcnunes:
Good idea @darkwisperer  and @BornaP :+1: 
I have thought bout it as well in the beginning of the project https://github.com/sqlectron/sqlectron-gui/issues/24. Some time later I closed that issue to keep the app more "minimalist".

But since there are people interest in this feature I would gladly accept a pull request for it :dancers:  

I prefer the mysql workbench approach. Seems more organized at least. Is important to also think in those DBs that have **scheme** such as PG (I guess oracle as well). Would be necessary keep all this structure for each scheme. Example:
[...]

PR #102, opened 2016-04-03
#98 show db metadata

BornaP:
Pull request regarding the #98 issue. 

I extracted table rendering from `database-list.jsx`. Tables, views, functions and stored procedures are now in a new component called `db-metadata-list.jsx`. At first I thought about having three separate components but than I realized there's a lot of shared code and it makes not much sense to keep them in separated files. 

By default, tables are shown when database loads, while others are not shown. All lists can be toggled (opened/closed) by a click. Component keeps information of opened/closed lists via its `state` object. Tables and views are clickable and by double-clicking on them default query is generated. For now, procedures and functions list items are not clickable, but maybe in future they could be - like for example showing SQL code on double clicking them. 

Of course, for this to work you have to use `sqlectron-core master` compiled, current version from npm won't work.

By working on this, I realized few more issues like when you click 'minus' icon to 'close' one database and double click another database, old ones remain 'open'. This is not a big bug but it's annoying when you want to change databases in the same connection. I opened new issue for this.
Also, when there are no tables in database nothing is shown. I think we should render metadata menu list containing a message that current db contains no tables or something like that (like here for views/functions and SP). Anyways, I'll open a new issue for this when PR is accepted, to keep this pull request clean from that.

Cheers,
Borna


src/renderer/components/db-metadata-list.jsx
```diff
@@ -0,0 +1,100 @@
...
+  }
+
+  toggleListCollapse(toggledList) {
+    const state = { ...this.state };
+    if (!state[toggledList]) {
+      state[toggledList] = true;
+    } else {
+      state[toggledList] = false;
+    }
+    this.setState(state);
```
maxcnunes: Could replace all this code with: 
```js
this.state({ [toggledList]: !this.state[toggledList] });
```

src/renderer/components/db-metadata-list.jsx
```diff
@@ -0,0 +1,100 @@
...
+    return listItems.map((item, idxChild) => {
+      if(listName === 'Tables' || listName === 'Views') {
+        return (
+          <span key={idxChild}
+            className="item list-item clickable"
+            onDoubleClick={() => onSelectTable(database, item)}>
+            {item.name}
+          </span>
+        );
+      } else {
```
maxcnunes: There is no need for `else` statement if the previous `if` is already leaving the function.

maxcnunes:
Awesome pull request!! I'm gonna test it locally before merging.

Related to the found bugs/improvements. I agree it is better create separated issues for them.

BornaP:
Thanks!

I agree completely with comments regarding the code. I'll make changes and push new commit!

src/renderer/components/db-metadata-list.jsx
BornaP: Are you sure about this one though? Seems better to keep it this way and use native setState method than mutating this.state directly.

maxcnunes: Sorry. I mean: `this.setState({ [toggledList]: !this.state[toggledList] });` then there is no need for `if/else`.

BornaP:
Changed! :)

maxcnunes:
Thanks @BornaP :clap:

BornaP:
You're welcome @maxcnunes !

Issue #97, opened 2016-03-15
Lost window when trying to connect to a DB with bad credentials


darkwisperer:
Client: MSSQL
APP version: 1.5.0
OS: Win7

While connected to a sql server, if you try to connect to a DB (from the left hand area) that your user does not have access to, you will be dropped out to the main screen and told that it was unable to connect.

Suggested handling of this: Instead of losing the entire window it should gracefully tell the user about the failed login.  Maybe flash red or highlight the DB in red or some means of marking that the connect won't work as is.

I was already connected to a DB with several tabs open when I chose to see how sqlectron would handle this edge case.

This issue might affect the other clients. I will try with mysql and report back.

maxcnunes:
Good catch!

I don't have much time to fix it right now. If you could send a pull request I would gladly accept it :smile:

darkwisperer:
If it is still open after I work on the other tickets I have agreed to I will :)

krolow:
have #106 solved that issue?

darkwisperer:
Sounds like it should. I was waiting for a new gui release to confirm this.

maxcnunes:
Sorry @darkwisperer. I was hoping release it earlier. But I have upgraded the build configuration. and I have not got it working properly yet. I hope finish it until this weekend.

darkwisperer:
@maxcnunes no rush :)
Just happy to help out when/were I can.

darkwisperer:
Issue still persists in v1.6.2

darkwisperer:
Still persists in v1.13.0

darkwisperer:
Can confirm still an issue in 1.29

PR #106, opened 2016-04-06
Fixes #105

### LINKED ISSUE #105: Can't connect to the other connection when there is one refused in the same session
Bug. To reproduce it:

1. Try connecting to saved connection which will return connection error (I used connection with PostgreSQL db here). Before connecting make sure that db engine is shutdown so you can get an error.
2. After error is shown try connecting to other saved connection which should connect successfully (I used connection with MySQL db)
3. Instead of establishing connection (with MySQL db), error with parameters from previous connection request is returned

BornaP:
Resolved issue #105 . Thing was that `serverSession` was still hanging on unsuccessful connection, so if you switched between different servers and previous one had an error connection, same error was returned.

### DIFF
--- a/src/renderer/actions/connections.js
+++ b/src/renderer/actions/connections.js
@@ -79,7 +79,10 @@ export function connect (id, databaseName, reconnecting = false) {
       if (dbConn) {
         dbConn.disconnect();
       }
-      this.disconnect();
+      const currentConn = getState().connections;
+      if (!currentConn.databases.length) {
+        this.disconnect();
+      }
     }
   };
 }
--- a/src/renderer/actions/connections.js
+++ b/src/renderer/actions/connections.js
@@ -79,6 +79,7 @@ export function connect (id, databaseName, reconnecting = false) {
       if (dbConn) {
         dbConn.disconnect();
       }
+      this.disconnect();
     }
   };
 }


maxcnunes:
Your change may cause problem in another situation. For example: I am already connected to a database. And I try to connect into another database in the same server. If this new connection crashes, it will close all the other connections as well - https://github.com/sqlectron/sqlectron-core/blob/master/src/db/server.js#L36.

Maybe a solution is checking if there is only one database "connected" before closing the server session.

darkwisperer:
in regards to @maxcnunes's example, there is a related issue open for when the connection fails it closes all connections #97.

BornaP:
@maxcnunes I see, my bad, haven't thought about that scenario. I'll try to figure it out later today or tommorow when I catch some time. 

Maybe another soultion could be additional checking if [sessionServer](https://github.com/BornaP/sqlectron-gui/blob/105-error-connection/src/renderer/actions/connections.js#L41) fits `serverConfig ` in current scope.

maxcnunes:
The sessionServer handles all connections to a same saver. Since is not possible to connect into multiple servers at same time. Should exists only one sessionServer per time. So I don't think your suggestion would be the best approach.
I think, if a session connection fails we should not keep it in memory. It should be discarded. So is better improve the error handling then keep failed sessions consuming memory unnecessarily.

BornaP:
Updated. Session is now discarded if it fails to connect. At first I thought about using error.code to check if server connection fails, but than I realized there are several more use cases with severe bugs like when you try to connect to nonexistent DB (in mysql) and then try to connect to postgresql which is shut down, but has database with same name like one in mysql, it would connect to mysql server and entitle database list like it is PostgreSQL server. 

So I returned to first thing @maxcnunes proposed and on error checked if current connection is connected to any database. It works just fine.

maxcnunes:
Thanks @BornaP :clap:

PR #110, opened 2016-04-11
Fixes #109 - Query box now resizes well

### LINKED ISSUE #109: Query box does not resize well
If I open a database connection and past or type in the query box a query with few lines (e.g. 15 lines). And I pull down the bottom of the box to increase the height. The query box will not resize well.

BornaP:
Thing was that ace editor triggered its resize method automatically only on window resize so it had to be invoked  manually. It's called now on ResizeableBox resize.

Seems that `onResize` method wasn't available in previous `react-resizable` version so I updated `package.json` (I also tried using Resizeable directly but I had other troubles with that). So you'll have to do `npm install` for this to work.

### DIFF
--- a/package.json
+++ b/package.json
@@ -80,7 +80,7 @@
     "react-ace": "^2.8.0",
     "react-dom": "^0.14.3",
     "react-redux": "^4.0.1",
-    "react-resizable": "^1.0.1",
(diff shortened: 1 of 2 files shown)


maxcnunes:
Thanks :clap:

PR #113, opened 2016-04-12
Closes #96

### LINKED ISSUE #96: UI: Add connection requires DB when making a connection
DB client: MSSQL, Mysql
App version: 1.5.0 
OS: Win 7

For both MSSQL and MySQL (not familiar with Postgress), when adding a connection the DB field should be optional since the database is not required to make the connection. I'm not sure if this is doable with the [mssql](https://www.npmjs.com/package/mssql) package. I could not find in their README what parameters were mandatory and which were optional. I know if using SQL Server Management studio it is an optional item. With the [mysql](https://www.npmjs.com/package/mysql#connection-options) package it is an optional parameter.
the user can always add it later if that is their preference

BornaP:
Included defaultDatabase argument in [here](https://github.com/BornaP/sqlectron-gui/blob/1140a45ff46e10b35b8d2b6fcf57bd32bdad1c35/src/renderer/actions/connections.js#L58). Now everything should work smoothly with this [commit](https://github.com/sqlectron/sqlectron-core/pull/11/commits/3e8b4ebbf54d85797a4bfb60009b53866cfbb55e) in sqlectron-core.

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,7 +48,7 @@
     "csv-stringify": "0.0.8",
     "debug": "^2.2.0",
     "react-paginate": "^0.5.0",
-    "sqlectron-core": "^3.1.0"
+    "sqlectron-core": "^3.2.0"
   },
   "devDependencies": {
     "autoprefixer-loader": "^3.1.0",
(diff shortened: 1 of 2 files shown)


maxcnunes:
@BornaP before I merge, could you include the new version o sqlectron-core in the package.json?

BornaP:
Sure, no problem!

BornaP:
@maxcnunes Again, new version of electron does not handle all errors properly. This time it does not return error on running query without addressing database or connecting to _PostgreSQL_ without existing default db. Instead of error being rendered it gets stuck in infinite loop. I'll degrade it back again since with 0.36.12 everything works fine.

maxcnunes:
Weird. I will try to remember testing this situation on upgrading the electron again. Thanks for the pull request @BornaP :clap:

PR #115, opened 2016-04-13
Fixes #111

### LINKED ISSUE #111: Table rows are not rendered in matching table headers in query result table
I just tried sqlectron with one bit bigger database (~130 tables) and executed default query on one of the tables (double-clicked table name). What I noticed is that all data is fetched correctly, but they are rendered under wrong table headers. They are mixed completely randomly. However, inspecting the elements I noticed that under data-reactid attribute correct column was set.

Here's example:

sqlectron_result

These attributes in <> (<name> ...) should be under column named `contentobject_name` , but they are rendered under `sort_order`. Element attribute `data-reactid` seems to have right column at the end.

BornaP:
Spent more time debugging this than fixing the code :smile: 

Thing was that attributes in JSON can be mixed up because it is, by definition, unordered structure. I still don't know the reason why this happened when there was more than 15 attributes in our case, but it did. So instead of depending on keys order in JSON (using `Object.keys()`), I changed it to use fields array received as a function parameter. It works now as expected!

### DIFF
--- a/src/renderer/components/query-result-table.jsx
+++ b/src/renderer/components/query-result-table.jsx
@@ -52,10 +52,9 @@ export default class QueryResultTable extends Component {
     }
 
     return rowsValuesToString(rows).map((row, index) => {
-      const columnNames = Object.keys(row);
       return (
         <tr key={index}>
-          {columnNames.map(name => {
+          {fields.map(({ name }) => {
             return <td key={name}>{row[name]}</td>;
           })}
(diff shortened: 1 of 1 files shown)


maxcnunes:
That was a weird bug. I would not have thought that. Thanks @BornaP :clap:

BornaP:
Yeah it was, took me a while to figure it out. You're welcome! :)

PR #118, opened 2016-04-14
Add default port when choose database

krolow:
This feature auto fill the port when choose one database...

We can not merge it yet as it depends of generate a tag for https://github.com/sqlectron/sqlectron-core/pull/12

I'll just leave it open to review code and once we generate a tag for  sqlectron-core, i can bump the https://github.com/sqlectron/sqlectron-gui/blob/master/package.json#L51 dependency

### DIFF
--- a/src/renderer/components/server-modal-form.jsx
+++ b/src/renderer/components/server-modal-form.jsx
@@ -71,7 +71,16 @@ export default class ServerModalForm extends Component {
   }
 
   onSaveClick() {
-    this.props.onSaveClick(this.mapStateToServer(this.state));
+    const data = {...this.state};
+    if (data.defaultPort) {
+      if (data.port === undefined) {
+        data.port = data.defaultPort;
+      }
+
+      delete data.defaultPort;
+    }
+
+    this.props.onSaveClick(this.mapStateToServer(data));
   }
 
   onRemoveCancelClick() {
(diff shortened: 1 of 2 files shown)


src/renderer/components/server-modal-form.jsx
```diff
@@ -130,6 +131,15 @@ export default class ServerModalForm extends Component {
     return hasError ? 'error' : '';
   }
 
+  handleOnClientChange(client) {
+    this.setState({ client });
+
+    const clientConfig = CLIENTS.find(entry => entry.value === client);
+    if (clientConfig && clientConfig.defaultPort) {
```
maxcnunes: Shouldn't it check if the port field has a value already?

krolow: i thought about that... i have even added that... but imagine that you are editing one connection you have added mysql but actually it was pgsql... should we disallow default port?

for me it makes sense too to not change when have state but while i was testing i found it a bit weird...

src/renderer/components/server-modal-form.jsx
```diff
@@ -232,7 +242,7 @@ export default class ServerModalForm extends Component {
                     name="port"
                     maxLength="5"
                     placeholder="Port"
-                    value={this.state.port}
+                    value={this.state.port || this.state.defaultPort}
```
krolow: @maxcnunes so i have change here so only if user has not choose any port we gonna apply default one

colecmc:
SQLectron is an awesome product. Thanks for sharing it.

I am using Version 1.21.0 (1.21.0.451) on a Mac. And I cannot connect to a Microsoft SQL DB because a default port is being added.
When I remove the default port, I am able to test the connection successfully, however I'm not able to save it.

What am I doing wrong? Please help.

eMerzh:
@colecmc could you open an separate issue about that ?  and not sure it will change but updating to the latest version could help too ;)

Thanks

colecmc:
OK, I've opened [334](https://github.com/sqlectron/sqlectron-gui/issues/334)

PR #119, opened 2016-04-15
Resolves #114 - Add db metadata filter

### LINKED ISSUE #114: Filter database metadata items
I have some databases that have too many tables. Would be awesome if was possible to filter the metadata items.

BornaP:
I implemented [option B](https://github.com/sqlectron/sqlectron-gui/issues/114) @maxcnunes was talking about.

Idea was to reuse the same component used for database filtering. Tested it locally and it seems to work fine.

Up for comments, especially regarding styling it :)

### DIFF
--- a/src/renderer/components/database-list-item.jsx
+++ b/src/renderer/components/database-list-item.jsx
@@ -1,5 +1,6 @@
 import React, { Component, PropTypes } from 'react';
 import DatabaseListItemMetatada from './database-list-item-metadata.jsx';
+import DatabaseFilter from './database-filter.jsx';
 
 
 const STYLE = {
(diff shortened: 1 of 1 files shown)


maxcnunes:
The code looks good for merge. Could you please provide a screenshot then I can evaluate the layout as well?

BornaP:
Sure thing!

filter_example

Basically, it's styled the same as the filter for databases (well it is the same component at the end haha).

maxcnunes:
I think that is all right. Not sure if will be a bit annoying the filter box always visible for every database.
But we can think better about it let this possible change to another issue.

maxcnunes:
Thanks @BornaP :+1:

BornaP:
Yeah, I agree with you. 
Again, you're welcome, glad I can contribute! \o/

PR #124, opened 2016-04-21
Table columns preview

BornaP:
Okay, this is still a working version of resolving [issue](https://github.com/sqlectron/sqlectron-core/issues/13) in GUI and is open for discussion.

Columns are separated in their own action/reducer and are loaded on (single) clicking table name. Columns are stored in a store object with db > table > columns hierarchy so they don't get mixed up if different databases have same tables.

Right now only columns are displayed and they aren't rendered nicely yet. This is where I need some advice regarding best approach, should I create new sub menu component that would contain columns and triggers or use current one somehow. Anyhow, component should be dynamically loaded as we fetch columns (and will do the same for triggers) on click. I guess once this is resolved, rendering triggers will be trivial (and probably the same as columns).

Well, I'm all ears!


src/renderer/actions/columns.js
```diff
@@ -0,0 +1,39 @@
...
+export function fetchTableColumns (database, table) {
+  return (dispatch, getState) => {
+    if (_shouldFetchTableColumns(getState(), database, table)) {
+      return dispatch(_fetchTableColumns(database, table));
+    }
+  };
+}
+
+
+function _shouldFetchTableColumns (state, database, table) {
```
maxcnunes: We don't use this standard of underscore for private functions. Could you please change it?

src/renderer/components/database-list-item-metadata.jsx
```diff
@@ -62,24 +63,51 @@ export default class DbMetadataList extends Component {
...
 
       const cssStyle = {...STYLE.item};
       if (this.state.collapsed) {
         cssStyle.display = 'none';
       }
+
       return (
         <span
           key={item.name}
           title={title}
```
maxcnunes: Running the app I realized the title "Click twice to select..." is showing even when I am with the mouse over the columns name. Should be good change the title now to appear only over the table name.

src/renderer/components/database-list-item-metadata.jsx
```diff
@@ -62,24 +63,51 @@ export default class DbMetadataList extends Component {
...
           {item.name}
+          <div className="menu" style={STYLE.menu}>
+            {this.renderSubItems(item.name)}
+          </div>
+        </span>
+      );
+    });
+  }
+
+  renderSubItems(_table) {
```
maxcnunes: Please do not use underscore for private variables.

src/renderer/components/database-list-item-metadata.jsx
```diff
@@ -62,24 +63,51 @@ export default class DbMetadataList extends Component {
...
+    const { columnsByTable } = this.props;
+
+    if (!columnsByTable || !columnsByTable[_table]) {
+      return null;
+    }
+
+    return columnsByTable[_table].map(column => {
+      return (
```
maxcnunes: No need for return in this case. Could do:
```js
.map(column => ({
  ...
}))
```

maxcnunes:
Your approach fetching the columns look good.
Related to appearance, I guess is everything too similar. Some icons could help on that:

Before:
screen shot 2016-04-21 at 12 17 06 pm
After
screen shot 2016-04-21 at 12 32 17 pm

```
<i className="table icon" style={{float: 'left', margin: '0 0.3em 0 0'}}></i>
<i className="columns icon" style={{float: 'left', margin: '0 0.3em 0 0'}}></i>
```

I guess the current behaivour to show the columns is a hidden feature. As a user I don't know if I would know is possible to list the columns. Maybe a title in the table could help (e.g. "Click once to... Click twice to..."). Is a bit ugly but at least is more visible. What do you think?

BornaP:
I like the idea with icons! Made both syntactic code changes (column actions now follows structure of similar action files) you proposed and added the icons. It looks much better now!

Regarding show user if it is clickable or not. I thought about adding collapse icon next to tables and on collapse sub menu with two headers would show up. On clicking those headers, columns/triggers should show. That's why I asked about how to resolve this, by making new component or reusing old one.
So something like this:
 ....
v Table_1
   v Columns 
         id
         col1
         ...
    > Triggers

Also, I'm aware that there is a bug that when you double click a table it performs a standard query and fetches columns. That's why I thought maybe standard query should be moved in right click menu (as I searched it is something doable in electron) or some other solution. Furthermore, maybe double clicking column should perform standard query only on that attribute?

Well that's just my 2 cents, what do you think about this? :)

maxcnunes:
Your suggestion of using collapse icon and submenus for columns and triggers looks good. Although I am not sure is really necessary move them to a separated component. Try out what looks better.

I agree moving the standard select query to a context menu. Actually @krolow  was talking with me this week about using context menus to do execute some few standard queries over the table, such as: select, insert, delete. So if you move the standard select query to a context menu is already a good progress for that feature.

I guess on MySQL workbench clicking over a column will just include the column name in the query box. Not sure what is best solution. I guess yours is more useful. But I gonna let you decide this one 😄

BornaP:
Sounds good :) Regarding MySQL Workbench, I've version 6.2 and there clicking over a column does nothing, while in context menu there is option to execute standard query. Since at the moment we don't (or even won't) have any more actions over specific column, I guess executing standard query with only that column included would be fine. Maybe later there could be included some kind of query builder where user could select more columns..but later on that, that would be a feature of its own :smile: 

PS. We've got 300th star on sqlectron-gui repo! :smile:

BornaP:
Updated, each table now has its own submenus. At the moment there's only submenu for columns, next thing is doing the same thing for triggers. Also, hovering over tables changes mouse pointer so users will know its clickable. However, it would be nice to make each table collapseable too.
Removed unnecessary table icons next to views, functions and prodecures.

I moved table submenu in separate component because of easier handling collapsibles. Secondly, whole rendering is specific to table so I think it's cleaner this way.

BornaP:
@maxcnunes I added table triggers and made whole menu collapsible. Now tables can be collapsed as well. Here's how it looks:

tablesubmenu

As far as I tested it, clicking again on table won't make new request for fetching columns/triggers (if they are already fetched), it will just collapse/expand it, which is good.

I guess this is it for this PR. I suggest moving context menu in separate issue/PR, to keep it cleaner because that's feature of its own. 
I'll leave it to you to review and test it locally and , if it is good enough, merge it. :)

src/renderer/components/table-submenu.jsx
```diff
@@ -0,0 +1,98 @@
...
+      <i className="columns icon" style={{float: 'left', margin: '0 0.3em 0 0'}}></i>
+    );
+
+    return itemsByTable[table].map(item => (
+      <span
+        key={item.name}
+        title={item.name}
+        style={cssStyle}
+        className="item">
+        {this.props.title === 'Columns' ? columnsIcon : null}
```
maxcnunes: I hope we find a icon for views as well in the future. This we keep the same pattern for both types.

maxcnunes:
That is awesome. The only thing I think could be improved is the typography for the columns. I think the columns looks a bit too small and thin. But I guess that is something we can release this way and see what the other users think about that.

By the way it is a great feature and your code style is pretty easy to read. Congrats!!

Again, thanks for the PR 👏 👏 👏

BornaP:
Thanks Max, you're very welcome! I agree on typography thou, but as you said when we get feedback we can change that :)

PR #128, opened 2016-04-27
Add query autocompletion

BornaP:
So I took some time today to figure this out. First off, if we do not plan to use autocomplete, react-ace =< 2.8.0 should be used, because upgrading to ^3.x.x causes some [problems](https://github.com/securingsincity/react-ace/issues/95). After some debugging and fixing unpredicted behavior, here it is - simple but neat autocompletion! It completes SQL keyword and words user recently typed (that would be column and table names in most of the cases I guess).

Thou adding autocomplete was a few lines of code, there were some unexpected behavior issues. `onChange` event on AceEditor now fires more often although string remains the same, which caused infinite number of `UPDATE_QUERY` actions. That's why I had to add [this](https://github.com/BornaP/sqlectron-gui/commit/55f5894a0f261cfcf3a544a3df30da016bf4e1be#diff-236a63be919fef1cdbb6ed49da761adeR69).  For some reason, I had difficulties with react-ace v3.4.0, autocomplete only worked for first word and was closing automatically after new `UPDATE_QUERY` action.  It works fine with 3.3.0 so that's why it's listed [explicitly](https://github.com/BornaP/sqlectron-gui/commit/55f5894a0f261cfcf3a544a3df30da016bf4e1be#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R70).

Before this is merged, I'll try if array of items (tables, columns) can be injected in autocompletion list. @maxcnunes @krolow please test it locally if it works on your machines as well! :smiley: 

Btw. Seem's that guys from WagonHQ (commercial version of our app :P) have extended ace with their custom autocomplete. [Link](https://www.wagonhq.com/blog/engineering-at-wagon)

### DIFF
--- a/package.json
+++ b/package.json
@@ -67,7 +67,7 @@
     "minimist": "^1.2.0",
     "node-sass": "^3.4.2",
     "react": "^0.14.8",
-    "react-ace": "^2.8.0",
+    "react-ace": "3.3.0",
     "react-addons-create-fragment": "^0.14.8",
     "react-dom": "^0.14.3",
     "react-paginate": "^0.5.0",
--- a/src/renderer/actions/queries.js
+++ b/src/renderer/actions/queries.js
@@ -66,9 +66,24 @@ export function executeDefaultSelectQueryIfNeeded (database, table) {
   };
 }
 
+export function updateQueryIfNeeded (query) {
+  return (dispatch, getState) => {
+    if (shouldUpdateQuery(query, getState())) {
+      return dispatch(updateQuery(query));
+    }
+  };
+}
 
-export function updateQuery (query) {
+function updateQuery (query) {
   return { type: UPDATE_QUERY, query };
+}
+
+function shouldUpdateQuery (query, state) {
+  const currentQuery = getCurrentQuery(state);
+  if (!currentQuery) return true;
+  if (currentQuery.isExecuting) return false;
+  if (query === currentQuery.query) return false;
+  return true;
 }
 
 
--- a/src/renderer/components/query.jsx
+++ b/src/renderer/components/query.jsx
@@ -3,6 +3,7 @@ import { debounce } from 'lodash';
 import AceEditor from 'react-ace';
 import 'brace/mode/sql';
 import 'brace/theme/github';
+import 'brace/ext/language_tools';
 import QueryResult from './query-result.jsx';
 import ServerDBClientInfoModal from './server-db-client-info-modal.jsx';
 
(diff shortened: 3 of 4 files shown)


maxcnunes:
I have tested it and it is working pretty well. For some reason ace does not have support for all common SQL keywords such as `LIKE`. But seems have support for most of them.
We can take a look in the future on react-ace to check if that problem have been fixed. For while I think is not a problem use a fixed version for react-ace.
Let me know if you need help including table names and columns in the autocomplete. I can take a  look on that as well. Maybe we will need follow the same approach of WagnonHQ extending the ace editor. But lets keep it simple if possible.

Also, I think maybe not everyone likes auto complete. Me for instance, I like them some times, but most of time I feel them a bit annoying. Could have an option to disable/enable the auto complete. Could be a configuration from the config file or a shortcut to toggle it. Also would be cool if was possible to disable the auto complete and only show the `suggestion keyword` on typing `CTRL+SPACE` or something like that. But these are changes we have to discuss more about. Lets keep for another issue/PR.

Thanks again for the PR 👏 👏 👏

BornaP:
@maxcnunes Great! Regarding including table names and columns in the autocomplete, we will probably just have to [write our own completer](https://github.com/securingsincity/react-ace/issues/69) which would contain tables/columns by database. If you want you can take a look, why not :)

And for keeping options in configuration, thats easly doable as what you have describe fits `enableBasicAutocompletion `option, while automatic autocomplete fits `enableLiveAutocompletion`. So I guess we could keep those bool vals in config.

Thanks for the merge! :tada:

maxcnunes:
@BornaP I will have some free time only in the weekend. Since you have found the solution and looks like be pretty simple to implement. If you are free to work on this and is excited to work on that, please do. Otherwise let me know than I can work on that in the weekend.

BornaP:
Sure thing, I'll try to get this done until the weekend

BornaP:
@maxcnunes I caught some little time for this and couldn't get it working today. I'll be off over the weekend so I'll get back to this later (if you won't resolve it in meantime). 
I tried to write custom completer following link earlier mentioned (which would include tables related to current database connection). I sent database name (`connections.databases[0]`) and `tables.itemsByDatabase` object as props to Query component, but problem was that database name is set to undefined at first moment for some reason.

maxcnunes:
@BornaP I have implemented the auto completion list with the database metadata ([commit](https://github.com/sqlectron/sqlectron-gui/commit/347b5c1ee28be0ca2de884ae24b7aadab40f34b7)).
But I found out sometimes while I'm typing the query it is jumping the cursor to end of the query.
I hope fix that tomorrow.

PR #130, opened 2016-04-29
Execute the selected query

rodrigowirth:


rodrigowirth:
image

maxcnunes:
Pretty handful feature. Thanks for the PR @rodrigowirth 👏 👏 👏 
There is only one problem. Your implementation does not work on executing the query through the short key. But no worries, I can fix that 😄

Issue #138, opened 2016-05-09
When press Ctrl + S tab get renamed to undefined


krolow:
![image](https://cloud.githubusercontent.com/assets/46822/15125835/54d35106-1605-11e6-8171-d3bb17e6a4fa.png)

BornaP:
Closed with 6a08e01a23ff023fa73d913ba72665393e58db73

PR #139, opened 2016-05-09
Fix when click in cancel while saving a query

krolow:
It's about #138 when click in cancel in save, now it's sends the action saying that there is a error...


krolow:
Closes #138

Issue #137, opened 2016-05-09
Be able to resize font at query editor


krolow:
![image](https://cloud.githubusercontent.com/assets/46822/15120419/02db3a1c-15ec-11e6-8137-14e3413dd523.png)

Default font-size it's too small, it would be good to be able to resize the font (at least use ctrl + scroll, to increase the font-size)

BornaP:
Closed with 33548c93aa445639bb010f362c58396257554d67

PR #140, opened 2016-05-09
Add support to increase/decrease font-size

krolow:
Related to #137 it's add Ctrl + "+" Ctrl + "-" to increase/decrease font-size... I'll check if I can add command to scroll as well

### DIFF
--- a/src/renderer/components/query.jsx
+++ b/src/renderer/components/query.jsx
@@ -157,6 +157,34 @@ export default class Query extends Component {
     });
   }
 
+  getCommands () {
+    return [
+      {
+        name: 'increaseFontSize',
+        bindKey: 'Ctrl-=|Ctrl-+',
+        exec: function(editor) {
+          const size = parseInt(editor.getFontSize(), 10) || 12;
+          editor.setFontSize(size + 1);
+        },
+      },
+      {
+        name: 'decreaseFontSize',
+        bindKey: 'Ctrl+-|Ctrl-_',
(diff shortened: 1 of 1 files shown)


src/renderer/components/query.jsx
```diff
@@ -157,6 +157,34 @@ export default class Query extends Component {
     });
   }
 
+  getCommands () {
```
maxcnunes: Do you think these commands should be included in the menu as well?

krolow: I didn't think so as the rest of interface might not be affected by the change only the editor...

If we apply in menu, for me it might give the felling that whole app should increase the font, and it's not the case...

maxcnunes: Could have a separate menu for that. Which would not cause this misunderstanding. But I guess is better keep it simple for now.

krolow: for sure, it would also impact, if you press the command without the focus in the ace editor, it would need some actions to change some store state through reducers to enable increase the font...

if it worth the effort i can implement... but i'm not sure if it worth

BornaP: @krolow I thinks it's the best to leave it as it is. :+1: for PR :)

krolow:
Alright i'll leave it like it is for now, I could found a wait to use "scroll" to increase font by a simple way, so I think it can be a new PR to add that

krolow:
Closes #137

Issue #126, opened 2016-04-26
Table rows scaling


BornaP:
I noticed that because of some long text (nvarchar) attributes, table rows do not scale nicely, they become too big. For example:

![table-sqlectron](https://cloud.githubusercontent.com/assets/8470710/14811590/03f188a0-0b99-11e6-8a5a-115bbed076b8.png)

Trying the same example with other DB tools, I realized they mostly do not allow word breaks and instead limit column width to certain number of chars (somewhere between 20 and 30 I guess). After certain number of chars shown, '...' is appended. If someone wants to see whole string, column width can be resized manually (also double clicking the columns selects whole text). Here's screenshot from MySQL Workbench, but MSSQL Server managment studio works the same 

![table_workbench](https://cloud.githubusercontent.com/assets/8470710/14811740/ce08ae52-0b99-11e6-976d-b7b94afbb785.png)

I think second way of scaling long strings/texts is much more convenient, as it reduces need for scrolling as more rows are shown within the same window.

krolow:
I think it's good, i like the approach of open a modal with a full text view in case of double click what do you think?

BornaP:
I thought about just selecting the whole text (so user can copy it), but I think modal would be fine :)

PR #145, opened 2016-05-13
Fix table rows scaling long texts

BornaP:
Pull request that resolves issue #126 

Table rows now have always the same height as long texts are not broken into more lines anymore. After certain number of chars, '...' is appended instead of the rest of the text. I've put fixed `max-width ` to 220 pixels, because I found out that corresponds to the same maximum number of characters shown in table cell in MySQL Workbench (which I took as a example). That can be changed if someone has better suggestion.

How it looked before:
table-sqlectron

How it looks now:
fixed_table_cell_scaling

I haven't implemented modal that @krolow suggested in the issue, because @maxcnunes proposed using [FixedDataTable](https://facebook.github.io/fixed-data-table/) where columns can be resized so user would be able to see full text that way (that's how many others SQL client's work). Also, full text/string can be selected by simple double clicking the cell value. So in that perspective, opening modal on double click is redundant.

### DIFF
--- a/src/renderer/components/query-result-table.scss
+++ b/src/renderer/components/query-result-table.scss
@@ -19,4 +19,11 @@
       }
     }
   }
+
+  td {
+    max-width: 220px;
+    overflow: hidden;
+    text-overflow: ellipsis;
+    white-space: nowrap;
+  }
 }


krolow:
:+1:

maxcnunes:
The good think about the modal is that we could prettify the data there. So you could show in plain text and have options to show as JSON, XML or event HTML.
But as you said, your approach + FixedDataTable solves the main problem of #126. So I guess that modal is just a nice to have and we can move that idea for another issue.

PR #149, opened 2016-05-16
Add support to disable the live auto complete

maxcnunes:
I feel the live auto complete really annoying.
This add support to disable the live auto complete. Actually it even sets the live auto complete disabled by default. But I'm ok in changing this default value if everybody else prefers it enabled by default.


BornaP:
I'm for leaving it enabled by default until we get normal settings window. Else, new user maybe won't be familiar it even exists.

maxcnunes:
Sure. I will let it enabled by default then.

Issue #134, opened 2016-05-05
DBs and Tables are not updating after create


cperezabo:
When you run a create database or create table, the created object doesn't show at the sidebar, you have to reload the complete app!

BornaP:
I think there is no database gui client (i.e. MySQL Workbench, pgAdmin) which shows newly created database/table without manually clicking on reloading data, but I agree we can try to implement this. Or at least put 'refresh button' which would reload databases and tables (like other clients do)

maxcnunes:
Even the auto refresh looking a best solution. We can not implement that because there is not a properly way to detect the query type in nodejs. Regex is not a good option and the best solution would be using AST of the query (which there is not any good one available in nodejs - is the same problem described [here](https://github.com/sqlectron/sqlectron-gui/issues/129#issuecomment-216000135)).
So the current best and simple solution is including a refresh button.

darkwisperer:
Random thought, on sql execute IF SQL string contains keyword "Create" then run the refresh menu function. I might be oversimplifying the issue.

i.e
`if ( string.indexOf('create') > -1 ||  string.indexOf('Create') > -1)
      refreshLeftMenu();`

maxcnunes:
Sorry. In this case there is no simple solution. Your suggestions would not work with many situations 😢 
```sql
INSERT INTO mytable VALUES ('I have created something today');
```
or
```sql
-- query created at 11/09/2011
SELECT * FROM mytable;
```
or
```sql
SELECT id, name, createdAt FROM people;
```
or
```sql
SELECT * FROM creations;
```
These are only the few simple cases I have just thought about right now. I'm sure there are others even more complex that a string "search" or even Regex would not work. Only a AST of the query would give us the full understanding of the query to be able to "react" over the content.

darkwisperer:
Your right. My idea would have unnecessarily done a refresh. I noticed there are a few AST related npm packages that we might be able to leverage. This is a little over my head, so I'm just going to follow along :)

cperezabo:
@BornaP oh yes it is. I use Sequel Pro for OS X and it show stuff just after the query run out!

krolow:
I think to begin a simple solution would be to add a refresh icon in the database, so we fetch the data once again.

It will not be automatic, but at least would give the support to refresh, and it might be a feature useful for other scenarios (like multiple users creating view/tables in same database) it would allow you to refresh and get the new things that others are generating...

[...]

PR #156, opened 2016-05-18
Feature/auto refesh

maxcnunes:
It refreshes dbs and tables based in the executed query. Requested on issue #134.

OT: I am impressed how was easy to implement it in the interface. I guess this is a sign of a good architecture design: including new features are soft pluggable. 👯

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -3,8 +3,8 @@
   "version": "1.9.0",
   "description": "A simple and lightweight SQL client with cross database and platform support",
   "author": {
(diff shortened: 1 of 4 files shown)


BornaP:
Really glad to see it was that easy to implement it on GUI! As you said, it's indicator of good architecture design \o/

@maxcnunes :clap: :clap: :clap: Great feature, I know it took you some time writing _sql-query-identifier_ but it was worth it. At least, it can be said now that we have one single feature some other SQL gui clients don't haha.

PR #157, opened 2016-05-19
Add search shortcuts (resolves #152)

### LINKED ISSUE #152: Add shortcut for database search and metadata search
Would be pretty useful at least for me 😄

BornaP:
Search shortcuts added to menu:

| Windows/Linux                  | Mac                            | Action                         |
|:-------------------------------|:-------------------------------|:-------------------------------|
| Ctrl+F | Command+F | Focus database search |
| Alt+Ctrl+F | Alt+Command+F | Focus database object search |

`Alt+Ctrl+F` focuses search on database that is currently active on query tab. If database connection is not opened or specified (i.e. user has lost focus from query area), command will be ignored. That's because current database name is needed so it can be resolved which search input to focus.

There's bug currently which I couldn't solve easily. When database search is focus, trying to focus database object search by keyboard does not react. I tried with setting all `toggle{Database}Search` states (expect the one triggered) to false, and act accordingly (i.e. on `nextProps.focusSerach === false` trigger blur, on `nextProps.focusSearch === true` trigger focus on input). But it did not work, it seemed like blur did nothing.

### DIFF
--- a/src/browser/menu/darwin.js
+++ b/src/browser/menu/darwin.js
@@ -135,6 +135,21 @@ export function buildTemplate(app, buildNewWindow, appConfig) {
       ],
     },
     {
+      label: 'Find',
+      submenu: [
+        {
+          label: 'Search databases',
+          accelerator: 'Cmd+F',
+          click: (item, win) => win.webContents.send('sqlectron:toggle-database-search'),
+        },
+        {
+          label: 'Search database objects',
+          accelerator: 'Alt+Cmd+F',
+          click: (item, win) => win.webContents.send('sqlectron:toggle-database-objects-search'),
+        }
+      ],
+    },
+    {
       label: 'Window',
       submenu: [
         {
@@ -174,4 +189,3 @@ export function buildTemplateDockMenu(app, buildNewWindow) {
     { label: 'New Window', click: () => buildNewWindow(app) },
   ];
 }
-
--- a/src/browser/menu/linux.js
+++ b/src/browser/menu/linux.js
@@ -103,6 +103,21 @@ export function buildTemplate(app, buildNewWindow, appConfig) {
       ],
     },
     {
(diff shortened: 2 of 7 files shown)


maxcnunes:
Is possible to get the current database with `this.props.queries.queriesById[this.props.queries.currentQueryId].database` even if the focus is not currently in the query box.

I think `F` is not the better letter to use as bind for this. If one day we implement search in the query box content I guess `F` seem like the perfect key binding for that. I tried find out another letter available, but seems most of them already have some function inside the query box. The character `.` does not have any function with `Ctrl`. I guess replacing `F` with `.` could work.

About the bug you described, I will try run your code locally to try find out a solution for that.

maxcnunes:
Trying to fix the bug you described I realized this approach does not really requires state changes.
Is much easier achieve that [exposing](https://facebook.github.io/react/tips/expose-component-functions.html) the focus function and executing it from outside of the component.

https://github.com/sqlectron/sqlectron-gui/commit/c986e8a771abb7658e66ddabfdf9558d76063f74

BornaP:
Yeah I agree with you, approach you proposed is much neater.

BornaP:
@maxcnunes I'll be off till Monday, so feel free to finish this if you want to :smile: I see you have implemented better approach so probably my commits for this PR are not needed.

maxcnunes:
I guess is just missing choose what bind key we gonna use for that.

BornaP:
`Ctrl+.` is fine if you ask me. Any other alternative I can think of is `F3` (I think it's used for search in some programs). Or perhaps `/` instead of `F` if it is not used for other shortcuts in editor. All three sound alright, so it's your call! :smile:

maxcnunes:
Close it since have been implemented here https://github.com/sqlectron/sqlectron-gui/pull/162.

PR #160, opened 2016-05-20
Add refresh database button

maxcnunes:
Fature discussed in #134.


BornaP:
Is this necessary if databases and tables are refreshed automatically on create/drop?

maxcnunes:
I guess there is two reasons for this button:

1. Allow refresh the database for those types the auto refresh does not have support yet (e.g. views).
2. Allow refresh the database in case someone else has created for instance a new table.

Issue #144, opened 2016-05-12
Replace paging result with "scrolling" result


maxcnunes:
After implementing #126 will be possible to use a "scrolling" result (because requires a fixed height for rows). Which I guess gives a much better UX. What do you think guys?

This one looks really good: https://facebook.github.io/fixed-data-table/.

BornaP:
I like the idea, big thumbs up for this! It looks much more 'native' for this kind of app. I'll try to implement #126 as soon as possible!

krolow:
it looks good this fixed data table

PR #164, opened 2016-05-23
Replace paging results with scroll

maxcnunes:
Issue #144.

wow. I never though would be this complicated to replace the paging with the scroll.
The `fixed-data-table` does not have features builtin for handling "automatic" size and resizing. At least the one they have (`flexGrow` I guess) is not exactly what we need. I had to implement a lot of stuff to give a better UX.
I guess is working well now. At least all the tests I have done seems to work.
Would be good if someone else good test it as well.

**OT:** Linking this PR with this issue https://github.com/facebook/fixed-data-table/issues/269, in case someone else needs an example of "automatic width of columns".


BornaP:
Good job @maxcnunes ! :)

BornaP:
I pulled this locally and it does not work as neatly as expected:

- At first I get only one row showed up in the table view and have to scroll for each following row. Later for some reason (probably because of resizing developers tools) it resized to 4 rows displayed at once. I'd change this to always have fixed number of N rows displayed in a view.

- There's an error I'm getting constantly:
query_result_table_warning

- I'd return ellipsis to the end of too long texts (it's css property removed in this commit).  And probably set max width of the cell to lower number of characters to reduce horizontal scrolling (I'd follow approach of Workbench and MS Studio here - I think both are limited to 220 px cell width).

maxcnunes:
* Here it shows all the available rows until the bottom of the window. Only if it has less rows to reach the bottom of window then the table will adapt to the height of the available rows.
* That error does not happen here. It used to happen on resizing the window. But that stopped when I added a debounce for the resizing.
* Initially I tried to keep the ellipsis. But didn't work in the first attempt then I just gave up. Since is possible to resize the column. I will try again a bit more. About the width I though was 220 letters. My bad, will change the width for pixels.

BornaP:
Cool cool cool. Just regarding 2nd point, are you sure you can't reproduce that error? Because sometimes I can't, but sometime it happens on resize. I guess this might be related to debounce time, maybe we should increase it a bit?

maxcnunes:
I'm trying to resize in different times and speed and it really does not happen. Lets merge to master with the current debounce time. Then you can change to a value the does work for you. Deal?

By the way, what OS are you using?

BornaP:
I'm okay with that, let's merge this then! :)

PS. I'm using Ubuntu 15.04

PR #169, opened 2016-05-24
Upgrade to electron stable major, v1.1.2

BornaP:
Finally, major version that works for us! :tada: 

Promises are now handled properly. There are bunch of warnings thrown in console but I guess they are related to Chrome v50.0.x, since I noticed them as of electron v1.1.0 which was bump to newer Chrome version. App works as expected anyways! :)

Since minor changes related to migrating to new API were already implemented in 131bf85154c3f976553a669f373695d58dc05049, this upgrade did not need any additional changes.

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,7 +54,7 @@
     "denodeify": "^1.2.1",
     "electron-builder": "next",
     "electron-packager": "^5.2.1",
-    "electron-prebuilt": "^0.36.12",
+    "electron-prebuilt": "^1.1.2",
     "eslint": "^1.7.3",
     "eslint-config-airbnb": "^0.1.0",
     "eslint-plugin-react": "^3.15.0",


BornaP:
Update: It seems there's Node v6 incompatibility with pg module we use. There's a problem with authentication, in short error saying that password is wrong keeps showing up, while password is actually alright. There's an [open issue](https://github.com/brianc/node-postgres/issues/1029). Since electron ~1.1 uses Node v6, I guess we hit another wall on updating to a major release.

maxcnunes:
Wow. Maybe one day we gonna be able to update to electorn 1.x 😆

BornaP:
Haha, somewhere in 2030 maybe. Well, I'll leave this open, because that issue might be solved soon so you can pull that changes to your fork we are pulling `pg` package from.

BornaP:
@maxcnunes There is actually already existing commit that solves this issue, so you call pull it on your fork. [Check it out](https://github.com/brianc/node-postgres/commit/6fddc566f09cf6b5919638a6f69c251e20437b75).

maxcnunes:
I have upgraded my pg branch. Since pg is a dependency based in a branch I guess there is no need releasing a new version of sqlectron-core. Just installing sqlectron-core should a get new version of pg module.

BornaP:
It's working :tada: :raised_hands:

maxcnunes:
Seems is not working properly. At least I cant connect over SSH. Could you check if it is working for you? Maybe we need revert again to `0.36.12` 😢

BornaP:
Ah, seems I've overlooked that :/ Could you please try debug it, and if it is not something that can be easily solved, we should revert this commit to 0.36.12 again.

maxcnunes:
The few minutes I tried investigating it I could not find the solution or even the exactly problem. Since I hope releasing a new version this week I guess is better revert it now. And later we try again upgrading it.
By the way I planning replacing the `remote` calls with HTTP requests. I hope after that change we will have less problems with electron.

BornaP:
Sounds good. Feel free to revert it :+1:

maxcnunes:
I guess this is the problem https://github.com/electron/electron/pull/5736. I am reverting to the previous used version. But lets try to upgrade again once that PR is merged.

Issue #180, opened 2016-06-01
Performance issues


ghedwards:
Performance issues / comments
- Query results view with large column set seems to perform badly, I've experimented with react-virtualized and slickgrid. Slickgrid seems to perform the best.
- Tables tree causes query editor to be quite laggy when there are a lot of tables.

BornaP:
Thanks for the feedback @ghedwards ! We are aware of the first point you made. Regarding the second one, could you explain a bit more, how much tables do you have? Do you find it laggy because of live auto complete? Because I did not find it to be laggy on databases with around 130 tables. (without using auto complete feature).

maxcnunes:
I have replaced the vanila HTML table with [fixed-data-table](https://github.com/facebook/fixed-data-table). I have not published yet because the build on CI is not working properly due http://stackoverflow.com/a/37559463/1050818. In the next release it should present a better performance.

ghedwards:
fixed-data-table seems to have performance issues as well and the project seems to be abandoned and not compatible with new versions of React? Slick grid is ok, but seems too to be a bit of a dead end, I've started looking at / coding https://bvaughn.github.io/react-virtualized/ as a replacement.

ghedwards:
1045 tables.. no further comment about the tables issue for now..

maxcnunes:
Not sure about the performance issues with fixed-data-table. But that is really not actively maintained over the last months. But what I have looked in react-virtualized I could not find support for grid with a header rows. That is why at the end I have used fixed-data-table.
Could wait until the next release to check if the performance has improved?

ghedwards:
100+ rows, 50+ columns.. scroll is laggy and slow. Running straight from master.

maxcnunes:
😢  Sorry, but I have just replaced the vanilla HTML table with fixed-data-table. I'm not really in the mood right now to rewrite it again with another library. Although a PR would be gladly accepted 😄

ghedwards:
Done, just need to clean up the code a little and make it look a bit nicer. Might have a PR in the next few hours or so.
<img width="712" alt="screen shot 2016-06-03 at 7 58 30 am" src="https://cloud.githubusercontent.com/assets/125586/15762412/feae7610-2960-11e6-8d31-86fdc6f5af5e.png">

maxcnunes:
Closed on https://github.com/sqlectron/sqlectron-gui/commit/678f7429f6d779cabbb6dcab3eb2f31a3bf90c04

PR #181, opened 2016-06-02
Implement react-virtualised query result table.

ghedwards:
#180


package.json
```diff
@@ -87,6 +87,8 @@
     "url-loader": "^0.5.7",
     "webpack": "1.12.14",
     "webpack-dev-middleware": "^1.2.0",
-    "webpack-dev-server": "^1.14.1"
+    "webpack-dev-server": "^1.14.1",
+    "react-virtualized": "^7.3.1",
+    "react-addons-shallow-compare": "^15.1.0"
```
maxcnunes: Seems is missing remove the  fixed-data-table from the dependencies.

src/renderer/actions/connections.js
```diff
@@ -61,12 +61,12 @@ export function connect (id, databaseName, reconnecting = false, sshPassphrase)
...
 
       if (!serverSession) {
-        if (server.ssh.privateKeyWithPassphrase && typeof sshPassphrase === 'undefined') {
+        if (server.ssh && server.ssh.privateKeyWithPassphrase && typeof sshPassphrase === 'undefined') {
           dispatch({ type: CONNECTION_REQUIRE_SSH_PASSWORD });
           return;
         }
 
-        if (server.ssh.privateKeyWithPassphrase) {
+        if (server.ssh && server.ssh.privateKeyWithPassphrase) {
```
maxcnunes: Seems this fixing is not related to this issue. Do you mind moving them to a separated PR?

src/renderer/components/query-result-table-virtualized-cell.jsx
```diff
@@ -0,0 +1,61 @@
...
+
+  render() {
+
+    const value = this.getValue();
+  
+    const className = classNames({
+      'ui mini grey label table-cell-type-null': value === null,
+    });
+
+    return (<div className="item" onContextMenu={::this.onContextMenu}>{value===null?<span className={className}>NULL</span>:valueToString(value)}</div>);
```
maxcnunes: This line is to large. Could move the component content to before the `return` or at least break the component content in multiple lines?

src/renderer/components/server-modal-form.jsx
```diff
@@ -61,7 +61,7 @@ export default class ServerModalForm extends Component {
       onUnchecked: () => this.setState({ ssl: false }),
     });
 
-    $(this.refs.privateKeyWithPassphrase).checkbox({
+    $(this.refs && this.refs.privateKeyWithPassphrase).checkbox({
```
maxcnunes: Can the refs be null or undefined at some time? Well, seems it is not related to this issue. If you could move it with the other one I suggested before would give us a better tracking in the repository history.

maxcnunes:
@ghedwards that is a great PR 👏 . I just suggested a few changes. I will try it locally later today.

src/renderer/actions/connections.js
eMerzh: Oups... didn't saw this pr before, and submitted a fix for that as PR #183

package.json
ghedwards: Thought I wouldn't remove it just yet in case someone wants to switch back to fixed-data-table ?

BornaP: There's no need to keep it under dependencies just in case someone wants to switch back. It's better to remove it from `package.json` and who really wants to use fixed-data-table can easly install it manually.

BornaP:
@ghedwards Good job, nice PR :clap: 

I'll leave it on Max to test it locally and merge it.

maxcnunes:
@ghedwards. I have tested it locally. Everything seems great but I have found 3 problems:

1. The table does not fit full window width:
screen shot 2016-06-04 at 3 27 45 pm
At least I think a full width looks better. For instance here is how was looking with fixed-data-table:
screen shot 2016-06-04 at 3 35 28 pm

1. There is no column resize and seems react-virtualized has no interesting implementing it into the library https://github.com/bvaughn/react-virtualized/issues/128. Well, now we have support to preview the full content of a column in a modal. But I'm not sure that is enough. A column resize would be really great.

1. When the there is no result is not showing any information. Previously it would show an empty table with row count = 0.

ghedwards:
Thanks, I'll take a look

ghedwards:
I've pushed a couple of updates to fix the width issue and the no records issue. For the resize I have some ideas on how this could be achieved but my wife has been admitted to hospital and so I might have limited time in the next couple of weeks. Please review the latest changes and let me know.

src/renderer/components/query-result-table-virtualized.jsx
```diff
@@ -234,30 +258,51 @@ export default class QueryResultTable extends Component {
...
-                    cellRenderer={this.renderHeaderCell}
-                    className="grid-header-row"
-                    rowHeight={30}
-                    rowCount={1}
-                    width={tableWidth - scrollbarSize()}
-                    scrollLeft={scrollLeft}
-                >
 
-                </Grid>
+                {(() => {
```
maxcnunes: Is there a reason for self invoking function here? If it really has reason for that I guess including some comment would help understanding it in the future.

BornaP:
First off, sorry to hear the bad news about your wife, hope she'll be okay. 

Secondly, no need to rush this PR. Feel free to take your time on resolving all issues.

maxcnunes:
I did a quick test. Seems the the table is resizing but is not quite good yet. As you can see the table get full width but with no background of the rows. Which gives the impression is not full width. 
screen shot 2016-06-10 at 7 30 56 am

By the way, I hope your wife gets better. As @BornaP said, no rush, take your time.

maxcnunes:
I have tested locally. The resizing looks pretty good. One last thing, the table height does not fit the number of rows. For instance, in case there are few rows it let blank a large part of the table.
Here is with your code:
screen shot 2016-06-14 at 7 51 40 am

Here is as it is in the 1.9.0 version:
screen shot 2016-06-14 at 7 51 49 am

src/renderer/components/query-result-table-virtualized.jsx
```diff
@@ -0,0 +1,388 @@
...
+    return <TableCell
+              rowIndex={params.rowIndex}
+              data={this.props.rows}
+              col={field.name}
+              onOpenPreviewClick={::this.onOpenPreviewClick}
+            />
+
+  }
+
+  handleStart() {
```
maxcnunes: Could we remove this empty function?

src/renderer/components/query-result-table-virtualized.jsx
```diff
@@ -0,0 +1,388 @@
...
+  }
+
+  handleDrag() {
```
maxcnunes: And this one too.

src/renderer/components/query-result-table-virtualized.jsx
```diff
@@ -0,0 +1,388 @@
...
+
+    for ( j=0; j < autoColumnWidths.length; j=j+1 ) {
+        totalColumnWidths = totalColumnWidths + autoColumnWidths[j];
+        if ( (j+1) === autoColumnWidths.length && totalColumnWidths < tableWidth ) {
+            totalColumnWidths = totalColumnWidths - autoColumnWidths[j];
+            autoColumnWidths[j] = (tableWidth - totalColumnWidths);
+        }
+    }
+
+    this.state.autoColumnWidths = autoColumnWidths;
```
maxcnunes: Is not recommend setting the state like this. Should use `this.setState({ autoColumnWidths })`. Also would give a better code readability if we could move the loop above to a function.

src/renderer/components/query-result-table-virtualized.jsx
```diff
@@ -0,0 +1,388 @@
...
+                    if ( fields.length > 0 ) {
+
+                        return <div style={{position:"relative"}}>
+
+                          <div className="scrollPlaceholder" 
+                            style={{left:(tableWidth - scrollbarSize()),
+                              width:scrollbarSize()}}></div>
+
+                          <Grid
+                              ref={(ref) => self.headerGrid = ref}
```
maxcnunes: I guess there is no need using self there since looks like you are only using arrow functions. So the context is always the one from the root function.

src/renderer/containers/app.css
```diff
@@ -40,3 +40,61 @@ span.list-item {
 span.clickable {
   cursor: pointer;
 }
+
+.Grid.grid-header-row .Grid__cell .draggable-handle {
```
maxcnunes: We usually keep the CSS styles in files related to the component.

maxcnunes:
Thanks @ghedwards. I have tested locally and it is looking really good. I have made some suggestions about the code. But don't worry, if you are tired of changing the code I can do them myself on merging the code into the master. Also I'm thinking in tweak a little bit the layout of the table to be a bit more similar with the rest of the application.

ghedwards:
Thanks @maxcnunes , I've got a bit on at the moment, I'm happy for you to make tweaks to the code / style where you see fit.

maxcnunes:
@ghedwards I have started working in your PR. I'm applying the project code style which is making me to change almost all the code (most of because of spaces, indentations, using vars instead of const and lets, and etc.). So after finishing it will looks like I implemented the feature instead of you. Because git blame will point the last change for me. To not take this feature away from you I am pushing this to a new [branch](https://github.com/sqlectron/sqlectron-gui/commits/react-virtualized). Then I can squash the commits I made into the yours and keep this way the credit of this feature in your name.

**Let me just make clear. Your code looks really good. Most of the changes are just code style.**

I believe this happened because the linter is not working well on `npm run test`. I`m going to fix that soon. Also I would suggest you to use the eslint plugin in your text editor. This help a lot during the development.

I wont be able to finish this today. Saturday I'm totally free for that. Then I can finish it and release a new version in the same day.

ghedwards:
@maxcnunes I'm all for code linting, in fact I use it a lot in my day job. I'm still learning ES6 / JSX, I'd be happy to run a linter over my changes in the future, how is best to do it in this project? I use sublime linter in my day job.

maxcnunes:
Cool. I have upgraded the eslint configuration. Now it is working properly. Although, seems there are almost 200 errors to fix. I'm going to fix everything on Saturday. After that running `npm test` will properly lint the whole project.

maxcnunes:
@ghedwards  Since would be a bit complicated to rebase it in a single commit and would lose a bit of the history. I have just squashed your PR with my changes into the master: https://github.com/sqlectron/sqlectron-gui/commit/678f7429f6d779cabbb6dcab3eb2f31a3bf90c04
Please take a look in the commits I had to do. I guess there are some suggestions you could take advantage of on developing in React.

Anyway, thanks a lot!!! 👏 👏 👏 👏

bvaughn:
Just saw this. Very cool! Thanks @ghedwards and co!

ghedwards:
Thanks @bvaughn , have a look at my other PR #200 , it needs some work and I don't have the time at the moment.

bvaughn:
That's funny! My next thought after seeing this PR was, "_I wonder if they've considering windowing the database/table explorer_".

I'd love to lend a hand if I can find some free cycles. Traveling in Japan this week so I'm not sure what my schedule will be like but I'll see what I can do.

PR #183, opened 2016-06-03
 🐛 fix: fix when connecting in local and trying to reach ssh settings

eMerzh:
When trying to connect while using the git version of sqelectron, 
i got the weird issue of

> Cannot read property 'privateKeyWithPassphrase' of undefined

here is a quick fix for that
using git , postgresql

mandatory gif :p 
mandatory gif

### DIFF
--- a/src/renderer/actions/connections.js
+++ b/src/renderer/actions/connections.js
@@ -62,7 +62,7 @@ export function connect (id, databaseName, reconnecting = false, sshPassphrase) 
 
       if (!serverSession) {
         if (server.ssh) {
-          if (server.ssh && server.ssh.privateKeyWithPassphrase && typeof sshPassphrase === 'undefined') {
+          if (server.ssh.privateKeyWithPassphrase && typeof sshPassphrase === 'undefined') {
             dispatch({ type: CONNECTION_REQUIRE_SSH_PASSWORD });
             return;
           }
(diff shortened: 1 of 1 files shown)


src/renderer/actions/connections.js
```diff
@@ -61,15 +61,16 @@ export function connect (id, databaseName, reconnecting = false, sshPassphrase)
       dispatch({ type: CONNECTION_REQUEST, server, database, reconnecting, isServerConnection: !databaseName });
 
       if (!serverSession) {
-        if (server.ssh.privateKeyWithPassphrase && typeof sshPassphrase === 'undefined') {
-          dispatch({ type: CONNECTION_REQUIRE_SSH_PASSWORD });
-          return;
+        if (server.ssh) {
```
BornaP: There is no need for this check if you also check it in nested IF statement. Could you please remove upper if and modify line 70 by adding `if ( server.ssh && ...)`. Like Gareth did in his PR before he removed it :)

BornaP:
@eMerzh Thanks for PR and fixing the bug! Will merge it after you make proposed changes :clap:

src/renderer/actions/connections.js
eMerzh: ok i've left the  if ( server.ssh && ...) from a previous test, 
but do you prefer a 
```js
if ( ssh ) {
if (a) {}
if (b) {}
}
```
or 
```js
if ( ssh && a ) {}
if ( ssh && b) {}
```

personally i think i prefer the first one ....

BornaP: I believe it's all the same in this case. You can leave it as you prefer. :)

Issue #165, opened 2016-05-23
Support for database diagram


BornaP:
Okay, so here's the idea which I'll try to implement (as it is one of features my mentor requests me to do, and I guess it would be really nice one to have!). I was thinking of making database diagramming component which would produce database visualization based on database tables, views and their references (primary and foreign keys! -> this will need some smaller extension of back-end API). The idea is to _spit out_ simple HTML tables connected with related tables. Furthermore, user would be able to reopen diagram once it's already generated (so some simple JSON with coordinates would be created upon generating diagram). I was searching around for suitable library and found [JointJS](http://www.jointjs.com/tutorial/html-elements) and [JS-graph-it](http://js-graph-it.sourceforge.net/index.html) to be probably the best choice for this use case. Both mentioned libraries have things we need - simple API, connecting custom HTML elements (which would be our tables/views), draggable nodes and connections. Of course, if anyone has better library to propose, please do.

**TL;DR:** I'll try to build support for creating database diagrams and storing them so user can reopen already existing diagrams.

**EDIT:** Think I'll go with `JointJS` as it is hosted on `npm`, supports nesting children elements inside of a parent (i.e. each table is a parent element with sub elements for each attribute), offers a rich API and supports graph export to JSON and reading it from JSON, so persisting the graph object should be trivial.

BornaP:
@maxcnunes @krolow @esnunes : All suggestions and opinions are welcome!

maxcnunes:
This would be an awesome feature. Not sure which library is the best for ERD diagramas. Seems like JointJS has a better documentation/support. I have found this GoJS (erd example http://gojs.net/latest/samples/entityRelationship.html). But I am not sure it is free.

BornaP:
GoJS is $1350 per developer :smile:  But no worries, I think JointJS has everything necessary to customly build simple ERD diagram. Will give it a try!

BornaP:
**UPDATE:** 
After including [latest changes](https://github.com/sqlectron/sqlectron-core/pull/21) in sqlectron-core, this is what I got so far:

![wip digram0 1](https://cloud.githubusercontent.com/assets/8470710/15711885/5ce67892-2810-11e6-9201-8d99d96887e9.png)

It's working on live data! (on picture it's some sample database I had locally)

[...]

PR #189, opened 2016-06-06
Database diagram feature

BornaP:
As discussed in #165 , implemented database diagram feature.

User can now right click on database and choose option '_Show Database Diagram_'. After that, user gets following modal:

opening_modal

Here user can select which tables (or views) he wants to include on diagram. After selecting at least one table, '_Generate diagram_' gets enabled and user can generate diagram containing selected tables. Spinner is showed up (in the same modal window) while columns and references are being fetched (if they weren't fetched earlier). When load is complete, diagram gets shown up. Here's and example:

diagram

Table elements on diagram are scaled based on longest string in table and number of table attributes. Links are rendered base on tables foreign key references. It's important to mention that tables are placed on diagram in very rudimentary way, but user is enabled to move elements as he likes. Smarter positioning elements on canvas would require more advanced algorithms for that. 
Also user can manipulate links between tables.

@maxcnunes I decided to separate persisting diagram on disk (and exporting it to svg/png) for another PR as this one got big enough itself. Also, since a lot of changes are happening recently it's probably smarter to divide it in couple of smaller PR's to avoid merge conflicts.

EDIT: In addition to save/open from file system feature, I plan to add key icons to Primary and Foreign keys (this requires update on back-end) and show up data types in future PRs.


maxcnunes:
Good job @BornaP. The diagram looks really great 👏 👏 👏 .

I am just confused about the user be able to manipulate links between tables. Since it is a diagram that does not apply change to database. Does make sense the user be able to do that?

By the way, since I have got the build working. I was planning in release a new version tomorrow. Do you think it is something we can include in the next release or should we wait for the following one?

src/renderer/components/database-diagram-modal.jsx
```diff
@@ -0,0 +1,156 @@
...
+      },
+    }).modal('show');
+  }
+
+  componentWillReceiveProps(nextProps) {
+    this.showDiagramIfNeeded(nextProps);
+  }
+
+  onSelectAllTables() {
+    $(':checkbox', 'div.ui.list').prop('checked', true);
```
maxcnunes: Do we really need JQuery to set values for the checkbox? Why not use only react for that?

src/renderer/components/database-diagram.jsx
```diff
@@ -0,0 +1,125 @@
+import React, { Component, PropTypes } from 'react';
+import joint from 'jointjs';
+
+require('./jointjs-diagram-table.js');
+require('./jointjs-diagram-table-cell.js');
+require('jointjs/dist/joint.min.css');
```
maxcnunes: Usually is better place the external styles before the internal customized styles. So they have "higher" rules.

src/renderer/components/database-diagram.jsx
```diff
@@ -0,0 +1,125 @@
...
+          newTabCell = new joint.shapes.sqlectron.TableCell({
+            position: {
+              x: (currentTable.position().x),
+              y: (currentTable.position().y + (idx + 1) * 20),
+            },
+            size: {
+              width: 100,
+              height: 20,
+            },
+            name: `${column.name}`,
```
maxcnunes: There is no need for template string here.

src/renderer/components/database-diagram.jsx
```diff
@@ -0,0 +1,125 @@
...
+            tableName: `${table}`,
```
maxcnunes: There is no need for template string here.

src/renderer/components/database-diagram.jsx
```diff
@@ -0,0 +1,125 @@
...
+              source: { id: currentTable.id },
+              target: { id: tableShapes[targetIndex].id },
+            });
+            newLink.attr({'.marker-target': { fill: 'yellow', d: 'M 10 0 L 0 5 L 10 10 z' }});
+            tableLinks.push(newLink);
+          }
+        });
+      });
+    } catch (e) {
+      console.log(e);
```
maxcnunes: In case a error occurs. Shouldn't it be displayed to the user? But if the console.log is really required. An approach that I would suggest is using a suffix in the log, such as `console.log('Error rendering diagram', err);`. Because in case an error comes as undefined because of some internal problem. At least we know is passing through this catch.

src/renderer/components/database-diagram.jsx
```diff
@@ -0,0 +1,125 @@
...
+    columnsByTable: PropTypes.object,
+    links: PropTypes.object,
+  }
+
+  constructor(props) {
+    super(props);
+    this.graph = new joint.dia.Graph();
+  }
+
+  componentDidMount() {
```
maxcnunes: I guess there is too much stuff going on in the componentDidMount. If you could break down in smaller functions would be really great. Maybe everything that you have included a comment can become a function.

src/renderer/components/jointjs-diagram-table-cell.js
```diff
@@ -0,0 +1,36 @@
...
+    this.updateCell();
+  },
+  render: function() {
+    joint.dia.ElementView.prototype.render.apply(this, arguments);
+    this.paper.$el.prepend(this.$box);
+    return this;
+  },
+  updateCell: function() {
+    const bbox = this.model.getBBox();
+    this.$box.css({ width: bbox.width, height: bbox.height, left: bbox.x, top: bbox.y, transform: 'rotate(' + (this.model.get('angle') || 0) + 'deg)' });
```
maxcnunes: This line is too long. Could you break down each property in a new line?

src/renderer/components/jointjs-diagram-table.js
```diff
@@ -0,0 +1,40 @@
...
+  },
+  render: function() {
+    joint.dia.ElementView.prototype.render.apply(this, arguments);
+    this.paper.$el.prepend(this.$box);
+    return this;
+  },
+  updateBox: function() {
+    // Set the position and dimension of the box so that it covers the JointJS element.
+    const bbox = this.model.getBBox();
+    this.$box.css({ width: bbox.width, height: bbox.height, left: bbox.x, top: bbox.y, transform: 'rotate(' + (this.model.get('angle') || 0) + 'deg)' });
```
maxcnunes: Break down this line :)

src/renderer/containers/query-browser.jsx
```diff
@@ -154,6 +157,31 @@ class QueryBrowserContainer extends Component {
...
+    dispatch(showDatabaseDiagram(database.name));
+  }
+
+  onShowDatabaseDiagram(database) {
+    const { dispatch } = this.props;
+    const selectedTables = [];
+
+    $(':checkbox:checked', 'div.ui.list').map((index, checkbox) => {
+      selectedTables.push(checkbox.id);
+    });
```
maxcnunes: This could be rewritten into this:
```js
const selectedTables = $(':checkbox:checked', 'div.ui.list')
  .map((index, checkbox) => checkbox.id);
```

maxcnunes: Also, is there a reason is using JQuery here instead of only react?

BornaP:
Thanks @maxcnunes ! 

Yeah you are right. User should not be able to disconnect links from elements. I'll (actually already have) changed that, but will leave option to user to bend connections on joints and move the place where link is positioned on the element - called port in JointJs (this two combine for very useful feature: if you have a lot of tables and don't want links to overlap). Something like this (links can only be bent and moved around the same element, not disconnected):
links_can_be_bent


I know you were planning on 1.9.0 for quite some time and I think it's better to make this release without this feature and then include it in the one after that so that users have more options on choosing which version to use :) Also, as mentioned before, I plan on adding some additional features to this diagram like saving so I guess when all that is complete we can publish it in its own release. TL; DR feel free to make a new release tomorrow without this :smile: 

I'll make changes you proposed in code!

src/renderer/components/database-diagram.jsx
BornaP: Hmm, so you wanted to suggest moving JointJS external styles to `app.jsx` or just above requiring jointjs-diagram elements?

src/renderer/containers/query-browser.jsx
BornaP: This is actually not the same. Your code would contain objects that jQuery returns. I'm using jQuery here because of easier reading checked checkboxes. I guess reading refs of the refs of the refs (react way) would be more complicated for only reading which checkbox is selected? Though I agree jQuery way might be replaced for react way of handling checkboxes in `database-diagram-modal.jsx`

src/renderer/components/database-diagram-modal.jsx
BornaP: After some thinking I think I'll left it as it is (using jQuery). Handling it the react way would require changing states which would trigger on change state component life cycle often, which is unnecessary in this case. Or am I missing something?

src/renderer/components/database-diagram.jsx
BornaP: I'll move styles specific to database diagram to separate .css file and load them after external styles.

src/renderer/components/database-diagram-modal.jsx
eMerzh: Yeah but you could also just use vanillaJs to do that no ?
not sure but you might want to look smth like document.querySelectorAll('div.ui.list [type="checkbox"]');

PR #192, opened 2016-06-09
Diagram persistence

BornaP:
What has been done:

- Added ability to persist diagram on file system and open it later. Diagrams are saved as a JSON file (native JointJS API). 

- Expanded initial diagram modal view with option for opening existing diagram.

- Extracted methods related to file manipulation (`saveFile`, `openFile`...) to `utils/file-handler.js`

How does it look:

Initial modal screen where user can choose between generating new diagram or opening existing one:
inital_modal_screen

Save button added at the bottom:
save_button

Possibly future enhancements:

- Add export to svg (and/or some picture format) option next to save option


maxcnunes:
Looks all good.

PR #194, opened 2016-06-12
Mark keys on diagram

BornaP:
* Upgraded Sqlectron to use `sqlectron-core v6.2.0`. Instead of previously used `getTableReferences`, referenced tables are fetched via `getTableKeys` method.

* Added key icon next to table keys. If column is PK, icon is yellow, else if column is FK, icon is black (default color). Makes diagram more informative :)

Screenshot: 
diagram_with_keys


maxcnunes:
Is not related to this PR. But I think it would looks better if this modal had a header :)

BornaP:
Okay, could do that. Don't know what to display in a header though, `${database_name} diagram`  or just 'Diagram' since user can select only some of the tables to display?

maxcnunes:
I guess just "Diagram" is enough.

BornaP:
I know this is not strictly related to this PR, but since it's a small change I pushed it here.

PR #198, opened 2016-06-15
Export diagram to image format - .png and .jpeg

BornaP:
Title says it all, PR implements feature enabling user to export diagram to `.png` and `.jpeg` format. :tada:

Next to save button, there's  simple dropdown button which offers user option to export diagram to one of the formats.

Although I thought it will be much simpler to export to SVG then png/jpeg, it turned out to be quite opposite. Reason is following. When I was creating diagram I followed JointJS [doc and tutorials](http://jointjs.com/tutorial/html-elements) on offical page. There is actually tutorial and example on how to build your custom elements containing plain HTML. Now it turns out, you can not simply do `svgImage = paper.svg` ,because `svg` JointJS spits out does not contain custom HTML elements. JointJS places them just above real basic `svg` elements (which I knew before but never thought it might be a problem :) ). So to enable export to `svg` it's required to rewrite diagram components to `svg` markup. At least I don't see any way around. After some searching around I found this [suggesting the same](https://groups.google.com/forum/#!msg/jointjs/QCeZ3ucyTNM/wcLkpnsOQ7MJ). Also there might be problem with loading external css. I'll try to catch up on this in a couple of weeks or so when I'll have more time. For now lets stick to 'image' formats.

Regarding the export to image formats, I used [html2canvas](https://www.npmjs.com/package/html2canvas) library which basically captures reconstructs 'screenshot' of a specified element. Quite neat tool actually :)

@maxcnunes Only one thing, there's a following `webpack` warning showing up after importing `html2canvas` :

webpack_warning_html2canvas

I guess there are two possible solutions - ignore it or fetch source directly from github? No?

### DIFF
--- a/src/renderer/actions/databases.js
+++ b/src/renderer/actions/databases.js
@@ -68,10 +68,6 @@ export function exportDatabaseDiagram(diagram, imageType) {
   return async (dispatch) => {
     dispatch({ type: EXPORT_DIAGRAM_REQUEST });
     try {
-      // fix - reapply css roles which html2canvas ignores for some reason
-      $('.link-tools, .marker-arrowheads', diagram).css({ display: 'none' });
-      $('.link, .connection', diagram).css({ fill: 'none' });
-
       // html2canvas library only captures elements in window view
       diagram.scrollIntoView();
 
--- a/src/renderer/actions/databases.js
+++ b/src/renderer/actions/databases.js
@@ -1,4 +1,5 @@
 import path from 'path';
+import html2canvas from 'html2canvas';
 import { getCurrentDBConn } from './connections';
 import * as FileHandler from '../utils/file-handler';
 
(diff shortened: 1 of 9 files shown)


src/renderer/actions/databases.js
```diff
@@ -60,6 +64,35 @@ export function saveDatabaseDiagram(diagramJSON) {
   };
 }
 
+export function exportDatabaseDiagram(diagram, imageType) {
+  return async (dispatch) => {
+    dispatch({ type: EXPORT_DIAGRAM_REQUEST });
+    try {
+      // fix - reapply css roles which html2canvas ignores for some reason
+      $('.link-tools, .marker-arrowheads', diagram).css({ display: 'none' });
+      $('.link, .connection', diagram).css({ fill: 'none' });
```
maxcnunes: Feels quite weird accessing the DOM inside the redux actions. Isn't this something you could do from where you are calling this function and just pass the result as argument to exportDatabaseDiagram? Maybe I'm wrong, but I have never seen this approach before.

maxcnunes:
Another indeed good feature. Thanks @BornaP 👏 👏 👏

src/renderer/actions/databases.js
BornaP: You're absolutely right! I was bit tired when writing this fix, so I forgot to move it. Good catch :)

BornaP:
You're welcome @maxcnunes . I made suggested code change, so If you are okay with npm warning, I'll merge changes into master.

PR #205, opened 2016-07-07
Doesn't show scrollbar on header any more, fix height for one or multiples rows

plandevida:
Changes:
- Fix height when result only has one row and the way to calculate multiple row height.
- Remove scrollbar from header table

Modified files:

- src/renderer/components/query-result-table.jsx
- src/renderer/components/query-result-table.scss

captura de pantalla de 2016-07-07 13-31-58
captura de pantalla de 2016-07-07 13-31-21

### DIFF
--- a/src/renderer/components/query-result-table.scss
+++ b/src/renderer/components/query-result-table.scss
@@ -47,7 +47,7 @@
 }
 
 .Grid.grid-header-row .Grid__cell {
-  overflow: visible;
+  //overflow: visible;
   font-weight: bold;
   padding-left: 10px;
   padding-right: 10px;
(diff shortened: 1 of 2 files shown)


src/renderer/components/query-result-table.scss
```diff
@@ -47,7 +47,7 @@
 }
 
 .Grid.grid-header-row .Grid__cell {
-  overflow: visible;
+  //overflow: visible;
```
maxcnunes: If is not required anymore then it could be removed.

maxcnunes:
Hi @plandevida, thanks for the PR. I'm going to test it this weekend.

PR #222, opened 2016-09-11
feat: display ssh configuration only if checked to unclutter the screen

eMerzh:
For now the connection management screen is quite clutter (for good reasons)
and things might not go by reducing the number of options available....
i propose to "hide" the ssh panel if the "ssh tunnel" switch is not active.

which cut down significantly the size of that panel

Before
capture d ecran 2016-09-11 a 23 25 18

After
capture d ecran 2016-09-11 a 23 25 09


maxcnunes:
Sure. That looks good. Thanks @eMerzh

PR #235, opened 2016-09-19
Fix connecting screen Test

eMerzh:
While trying the test connection i got some issues when not entering any informations
capture d ecran 2016-09-18 a 22 38 41

Here is a fix for that issue, not enabling the "Test" button if no client is selected.
in the meantime, i've also added a "clearable" = false on this field because it doesn't really make a sense to "clear" the client ... you can still change it of course :)

and preventing some error about uncontrolled to controlled fields

### DIFF
--- a/src/renderer/components/server-modal-form.jsx
+++ b/src/renderer/components/server-modal-form.jsx
@@ -189,12 +189,15 @@ export default class ServerModalForm extends Component {
 
   render() {
     const { testConnection } = this.props;
-    const { confirmingRemove, isNew } = this.state;
+    const { confirmingRemove, isNew, client } = this.state;
     const isSSHChecked = !!this.state.ssh;
     const ssh = this.state.ssh || {};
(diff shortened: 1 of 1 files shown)


src/renderer/components/server-modal-form.jsx
```diff
@@ -224,15 +227,16 @@ export default class ServerModalForm extends Component {
                   name="name"
                   maxLength="250"
                   placeholder="Name"
-                  value={this.state.name}
+                  value={this.state.name || ''}
                   onChange={::this.handleChange} />
               </div>
               <div className={`six wide field ${this.highlightError('client')}`}>
-                <label>Client</label>
+                <label>Database Type</label>
```
maxcnunes: Would be good if it had been included in a different commit and pull request.
But I will not make you rebase the PR just because of it. Just keep in mind next time to break down the pull request.

PR #237, opened 2016-09-21
feat: Add visual distinction to indicate if there are items below

eMerzh:
I was frustrated while browsing the database, 
stumbling upon many "No result found" after clicking the view submenu to open it.

So i thought about having some kind of visual distinction.. here is my take

Before:
capture d ecran 2016-09-19 a 23 19 27

After:
capture d ecran 2016-09-21 a 22 31 53

### DIFF
--- a/src/renderer/components/database-list-item-metadata.jsx
+++ b/src/renderer/components/database-list-item-metadata.jsx
@@ -39,15 +39,21 @@ export default class DbMetadataList extends Component {
   }
 
   renderHeader() {
+    const {
+      items,
+    } = this.props;
     const title = this.state.collapsed ? 'Expand' : 'Collapse';
     const cssClass = this.state.collapsed ? 'right' : 'down';
-
+    const cssStyle = { ...STYLE.header };
+    if (!items || !items.length) {
+      cssStyle.color = 'lightgray';
+    }
     return (
       <span
         title={title}
         className="header clickable"
         onClick={::this.toggleCollapse}
-        style={STYLE.header}>
+        style={cssStyle}>
         <i className={`${cssClass} triangle icon`}></i>
         {this.props.title}
       </span>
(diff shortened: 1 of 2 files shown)


src/renderer/components/table-submenu.jsx
```diff
@@ -33,15 +33,20 @@ export default class TableSubmenu extends Component {
   }
 
   renderSubMenuHeader() {
+    const { itemsByTable, table } = this.props;
+
     const title = this.state.collapsed ? 'Expand' : 'Collapse';
     const cssClass = this.state.collapsed ? 'right' : 'down';
-
+    const cssStyle = { ...STYLE.header };
+    if (! itemsByTable || !itemsByTable[table] || itemsByTable[table].length === 0) {
```
maxcnunes: Could you remove the space between the `!` and `itemsByTable`?

src/renderer/components/database-list-item-metadata.jsx
```diff
@@ -39,15 +39,21 @@ export default class DbMetadataList extends Component {
...
 
   renderHeader() {
+    const {
+      items,
+    } = this.props;
     const title = this.state.collapsed ? 'Expand' : 'Collapse';
     const cssClass = this.state.collapsed ? 'right' : 'down';
-
+    const cssStyle = { ...STYLE.header };
+    if (! items || !items.length) {
```
maxcnunes: Could you remove the space between the `!` and `items`?

src/renderer/components/table-submenu.jsx
maxcnunes: Since you gonna fix that. Could you also replace `itemsByTable[table].length === 0` with `!itemsByTable[table].length`? Is the way you usually use it in the project.

eMerzh: sure! :)
I thought about the length === 0, then I forgot /o\

PR #239, opened 2016-09-22
fix: substract table scroll from header to sync scrolls in results

### LINKED ISSUE #233: Misalignemnts in results header / data
i was doing a request with more columns than what fit in the screen and sqlectron gives me a weird alignment of the column names with the data ...

capture d ecran 2016-09-18 a 23 29 48

eMerzh:
here is an attempt to fix the scroll misalignment 
while having more rows than could fit on the screen, 

should fix #233 ... at least from here :)


maxcnunes:
I remember we had that code before. But I removed in some refactoring I guess because looked like it was not really applying any change. I hope it was there because of this problem 😄

PR #240, opened 2016-09-23
improve: Change disconnect icon to poweroff

eMerzh:
Hi, 

Here is a litte proposition with a little PR, 
i was really intrigued  by the choice of the "disconnect" icon, as semantic ui says, it was the "Ban" icon... which is not really what we want...

i tried the "sign-out" which is what feels more natural , but the look does not right to me : 
capture d ecran 2016-09-23 a 21 40 27

so i tried the universally understood "power off", and seems better, 
what do you think ?

capture d ecran 2016-09-23 a 21 36 21

Another question i had was about the reconnect icon; aside from the icon that don't feel really obvious to me either, isn't it possible to remove it ? or maybe put it only when we are disconnected or smth ?
what do you guys think?

### DIFF
--- a/src/renderer/components/header.jsx
+++ b/src/renderer/components/header.jsx
@@ -56,7 +56,7 @@ const Header = ({ items, onCloseConnectionClick, onReConnectionClick }) => {
             <button className="ui icon button"
               title="Close connection"
               onClick={onCloseConnectionClick}>
-              <i className="ban icon"></i>
+              <i className="power icon"></i>
             </button>
           </div>
         </div>


maxcnunes:
Looks good with the power off icon. 👏  Regarding the reconnect button. I have added that button because sometimes I lost the connection and the client didn't send any event about that. So it would get stuck and I would need to leave and connect again. Which would make lose the current tabs. That is way I have added that button. If we could detect better the lost connections that would be really better. But I am not sure is something so reliable. I guess MySQL has a button like that as well visible even if your connection is ok.

PR #241, opened 2016-09-24
feat: replace the database name by the Connection name with the DB logo

eMerzh:
I didn't noticed it at first, 
but i don't really understan Why the Type of DB is marked as the HEADER of my databases...

While it's a nice thing to know on which DB i am, i think it should have the name of the connection instead + the type of DB...
here is a screenshot : 
BEFORE
capture d ecran 2016-09-24 a 22 49 48

After:
capture d ecran 2016-09-24 a 22 43 25

i understand that it might make a duplicate with the Top bar, but for me it make more sense to have that here.


maxcnunes:
I usually work at same time in different databases. So, have the database type somewhere easy to see is really helpful.
In the beginning of the project I tried this approach (server name + db type icon) in the sidebar header. But I didn't like much how that looked. At the time we were not handling well long text in the sidebar. But nowadays looks good:
screen shot 2016-09-25 at 8 19 08 am

Although another problem still present. The icon for MySQL is not transparent. So, looks pretty ugly in a non white background.
screen shot 2016-09-25 at 8 17 47 am

> PG and Cassandra the icon looks good. MSSQL I have not tested.

I would gladly accept this PR once we have a better icon for MySQL. If is not possible to find a better icon. Maybe you could try a different approach such as moving the database type to the top header (Although maybe the icon will be too small and will look ugly as well :(, not sure).

eMerzh:
@maxcnunes , what do you think about this one https://en.wikipedia.org/wiki/File:MySQL.svg

maxcnunes:
Would be good to see how it looks in that background :)

eMerzh:
i've replaced the logo, 
it seems to be working well

capture d ecran 2016-09-25 a 15 38 36
capture d ecran 2016-09-25 a 15 37 46

PR #243, opened 2016-10-03
ux: hide tab mover (left/right) when the tab list fit on screen

eMerzh:
@maxcnunes  refactor the tab list that work really better now, 
but i think it "clutter" the screen a bit while you don't have many tab open...

here is a propostion about hidding arrow if the tab list fit in the screen
capture d ecran 2016-10-03 a 23 04 39

capture d ecran 2016-10-03 a 23 04 51

what do you think?


maxcnunes:
Pretty good. Thanks @eMerzh!!!

PR #244, opened 2016-10-03
[question] Add new tab button

eMerzh:
I think the way to create a new tab  is a bit hidden, 
as in mysql it's "recommanded" to run queries in distinct tabs, we should make it a bit more visible.

i tried to add a button in the tab list , that's where it feels the more natural to me, 
but i'm not really sure about the design or the placement, 

what do you think ?

capture d ecran 2016-10-03 a 23 29 40

### DIFF
--- a/src/renderer/containers/query-browser.jsx
+++ b/src/renderer/containers/query-browser.jsx
@@ -428,6 +428,9 @@ class QueryBrowserContainer extends Component {
               {menu}
             </TabList>
           </div>
+          <button className="ui basic icon button" onClick={() => this.newTab()}>
+            <i className="plus icon"></i>
+          </button>
           {isTabsFitOnScreen &&
             <button className="ui icon button"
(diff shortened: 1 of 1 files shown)


maxcnunes:
Without the nav buttons it looks good. But how does it look when the right button is visible?

eMerzh:
a bit weirder :s 

capture d ecran 2016-10-03 a 23 50 31

maxcnunes:
yeah, looks a bit weird. Maybe if the were like a composed button (without the radius border between them) would look better.

eMerzh:
by wrapping in in "buttons" i got
capture d ecran 2016-10-04 a 23 23 04
or
capture d ecran 2016-10-04 a 23 24 57


bit better but doesn't really feels natural; no?

maxcnunes:
I will accept the PR as it is and try to make it looks better. Thanks @eMerzh

eMerzh:
awesome ! thanks @maxcnunes

Issue #133, opened 2016-05-05
Opening Databases is not intuitive


cperezabo:
First of all (and after you discover how) it feels like if it would have a delay when you click on the database at the sidebar
But the most important thing is that the "+" for expanding the database tables isn't there by default!. You first have to discover that you have to double click the database name to open it, but just the name, because I was clicking the database box and nothing happened until I tried on the name. Then, the "+" appeared and the tables showed up.
The cursor pointer never changed to say me "hey, you can click or double click here"

maxcnunes:
At the beginning we had a pointer cursor on passing the mouse over the database. But we changed that to default cursor because you have to double click to open the database. And the pointer cursor is a common style for one click. The database name has a tooltip saying you can click twice to open the database. But that only appears if you stop the mouse over the database name.
But you are right, we have to make it more clearer for new users. I going to try out some stuff and see which goes better.
Thanks for the feedback.

cperezabo:
I think the first approach was better!. It is simpler to click once instead of twice.
And further more if you have to see a tooltip to know you can double click the database name!

BornaP:
I'm not sure if it is a good practice to require only a single click for opening new database connection, because user can click it by mistake and open new connection redundantly.

krolow:
I agree :+1: to have one click back and cursor pointer

cperezabo:
@BornaP At least is more intuitive!. UX is awful as is today. I double click a database and no "Opening..." message is displayed, further more I have to discover where to double click! in the name? in the surrounding area? in the db icon at the right? Opening a new connection on a "mistaken click" is paradise compared with the time it takes to discover where you have to click!

krolow:
yeah for sure @cperezabo, that's why we are welcome to PR and changes, it would be good to receive from you a PR to improve such behavior... just to get a context, we (when i say we, mainly @maxcnunes) we have built that because we haven't found a good App to query across platform in PgSQL, it was the begin of the idea of sqlectron...

so for sure it have a lot to improve as we are **developers** building something to use in our daily bases, it has a lot of UX and features to improve...

based on that is why we encourage others to use it, as well improve it by creating PR...

in the meanwhile it seems enough to only change it back to mouse "pointer cursor" and add one mouse click, what do you think?

maxcnunes:
I disagree with these options:
 * pointer cursor + one click: as @BornaP said it will make users open a new connection by mistake.
[...]

PR #250, opened 2016-10-06
feat: change double click on the DB header to simple click

eMerzh:
On the left sidebar, 
You have to Double click on the database Header name or reach the little arrow and simple click on it.
it feels unnatural for me, and the comportment is different in lower levels (tables, view, ...)

here is a proposition to make the thing work with a simple click.

### DIFF
--- a/src/renderer/components/database-list-item.jsx
+++ b/src/renderer/components/database-list-item.jsx
@@ -74,7 +74,7 @@ export default class DatabaseListItem extends Component {
     this.setState({ filter: value });
   }
 
-  onHeaderDoubleClick(database) {
+  onHeaderClick(database) {
     if (!this.isMetadataLoaded()) {
       this.props.onSelectDatabase(database);
       return;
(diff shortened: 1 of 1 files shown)


maxcnunes:
I guess this was discussed before in some old issue. My only concern using a single click is that the user open a new connection without really want to do that. By clicking by mistake over the database name. But if it is the way other sql client does that I'm ok to go with this solution.

eMerzh:
ok, so if i understand correctly, if you double click, we make a new connection, 
and you, somewhat, want to avoid that if not necessary. Right ?
Postico / pgcommander / MysqlWorkbench work with simple click to change db,
sequel pro use a "<select></select>" to change db , so not really relevant

eMerzh:
i guess you refere to this comment from @BornaP  ? https://github.com/sqlectron/sqlectron-gui/issues/133#issuecomment-217261931

maxcnunes:
Yeah. Is that issue I was talking about. Then was decided we were going to use the same way the Workbench does. https://github.com/sqlectron/sqlectron-gui/issues/133#issuecomment-217306706 Which is the way is working nowadays 😄 

But if the way is nowadays is really complicated to use I agree to discuss a new approach. I just would like to avoid the user opening a new connection by mistake. With remote connections on a slow network it can be a bit frustrating. Maybe including the icon can work. We just have to be careful to not put too much stuff in the interface. Maybe the icon be visible only on hover the database could work as well (but could be not so intuitive).

eMerzh:
ok i understand. but then the UI should be more restrictive, 

like for example, "ok let's browse one db a the time" with a select box or an accordion....
and that could help to have a better link between the tab's db and the browsed db...

maxcnunes:
Sorry, I could not imagine how the interface would look like with your description :laughing: 

I like the left sidebar with all dbs because it give fast access to all available dbs. Seems would be more complicated to work with different dbs at same time without that. But if you have a propose for a new interface that would improve the UX I'm open to discussion. Although, it looks a bit low priority based in other issues we have.

eMerzh:
Yes ok i'll close this... because the fix is more complex than the PR here ...

Issue #153, opened 2016-05-18
Add Support for schema


eMerzh:
With postgresql and other big databases (oracle, ...) you have schema...
it seems that sqlectron does not support them .... it would be really nice to be able to browse accross distinct schemas :)

BornaP:
Yeah you're right, sqlectron does not support them, it currently reads only public schema of certain database (if DB engine supports schemas). Since only PostgreSQL of all database that we support has schemas, I guess it would require writing some code specific only for postgres case, since the app is hybrid.

eMerzh:
ok... i understant that it's a bit harder than just "display_the_schema=yes"
but would'nt it be possible to at least let the user select a schema at connection time or something ?
or any other workaround ?

because Postgresql support without schema is not really as useful as it could be.

BornaP:
I agree with you on that. We'll see whether to implement schemas in database list so that user can browse between schemas (if they exists) or as you said, enable user to select schema when connecting to database.

I guess first option might be doable with making component that would be hidden when not using postgres (hence when method in back-end, let's call it _getDatabaseSchemas_ , would not return anything).

maxcnunes:
I guess this issue could be break down in 2 PR:

1. allowing the user select the default schema during the connection (in the server management screen).
2. listing the schemas in the database list as @BornaP said.

kstr:
I noticed that tables from another schema are simply not listed on the left side of the screen, but it is possible to include them in a query.  I miss this feature since we use a custom scema in all our databases.

abourget:
Splitting views per schema is a little tricky though, we have cross-schema queries and it'd be annoying not to have completion between schemas too... why not simply list the tables in the `Tables` view, prefixed with their `schema` ?

The search in there could help us navigate.. it does a great job at it.

maxcnunes:
Done on https://github.com/sqlectron/sqlectron-gui/releases/tag/v1.19.0

PR #257, opened 2016-10-13
Basic Schema support

eMerzh:
Hi, 

here is the first step for schema support.
Let's add a field in the connection screen , then use it in all selecting objects queries.

Then later we can work to add a bit of interactivity to select the schema.
i've also added the schema completion for the query.
but note: the custom query part is not affected by the schema "selection".

need https://github.com/sqlectron/sqlectron-core/pull/33

ref #153 
what do you guys think ?

### DIFF
--- a/src/renderer/actions/columns.js
+++ b/src/renderer/actions/columns.js
@@ -6,10 +6,10 @@ export const FETCH_COLUMNS_SUCCESS = 'FETCH_COLUMNS_SUCCESS';
 export const FETCH_COLUMNS_FAILURE = 'FETCH_COLUMNS_FAILURE';
 
 
-export function fetchTableColumnsIfNeeded (database, table) {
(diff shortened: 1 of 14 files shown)


iwozzy:
This will be super useful for connecting to Redshift. With the use of services like Segment, they import the data into Redshift using different schemas. 
Right now connecting and getting data out of Redshift is so painful and antiquated. Being able to use sqlectron will be such a breath of fresh air and a relief :)

simplesteph:
Any updates on merging this?

maxcnunes:
I got this month to close some issues. This one is in my todo list.

maxcnunes:
@eMerzh Thanks for the PR!! I have just merged it through another PR.

eMerzh:
awesome look forward to improve that :) 
and to keep moving sqelectron forward 🎉

PR #279, opened 2017-01-01
Issue 278 - change keybinding for window reload

### LINKED ISSUE #278: Change keybinding for window reload (⌘R) or change window reload behavior
Note: this is not a bug, and I would be happy to make a PR, but I wanted to see if anyone else had the same issue first.  This is more for an open discussion.

I came to Sqlectron from using a handful of DB clients together for different needs.  Some of those clients I used (Sequel Pro comes to mind) use ⌘R as the keybinding to run a query.  With this muscle-memory, I am constantly hitting the combination that maps to refresh the window.  This results in me watching the (very good) opening animation, seeing a disconnected session, and losing whatever query and past results I had open at the time.

I am proposing a couple changes:

  * changing the window reload keybinding to something other than ⌘R
  * have ⌘R do the same thing as ⌘⏎ (command-return)
  * change the window reload behavior to stop the "Server configuration not found" error (unless this is intended; in which case a different error message would be preferred)

Thanks in advance everyone!

blakek:
Hey everyone!  This is to resolve #278 

  - Keybinding to reload window is Ctrl/Cmd-Shift-R (to mimic behavior cache ignoring keybinding in Firefox and Chrome)
  - Ctrl/Cmd-R now also Executes query (keeping Ctrl/Cmd-Enter)

I may be wrong, but I don't thing multiple accelerators can be assigned in the Electron menus.  To get around this, I added an event listener for `browser-window-created` in `src/browser/app.js` that adds the duplicate shortcut to execute a query.

Any thoughts or changes needed?
Thanks!

### DIFF
--- a/docs/app/keyboard-shortcuts.md
+++ b/docs/app/keyboard-shortcuts.md
@@ -29,7 +29,7 @@
 | <kbd>Ctrl</kbd>-<kbd>T</kbd> | <kbd>Command</kbd>-<kbd>T</kbd> | New Tab |
 | <kbd>Ctrl</kbd>-<kbd>W</kbd> | <kbd>Command</kbd>-<kbd>W</kbd> | Close Tab |
 | <kbd>Ctrl</kbd>-<kbd>S</kbd> | <kbd>Command</kbd>-<kbd>S</kbd> | Save Query |
-| <kbd>Ctrl</kbd>-<kbd>Enter</kbd> | <kbd>Command</kbd>-<kbd>Enter</kbd> | Execute Query |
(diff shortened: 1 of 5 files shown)


eMerzh:
LGTM 👍 ...
and it also seems that you can't have 2 shortcut do the same thing by default in electron's menu.

maxcnunes:
@blakek Thanks for the PR. I will try to test that as soon I get a free time. @eMerzh If you have tested that already feel free to merge that.

eMerzh:
yep, work like expected. 
thanks @blakek

blakek:
Thanks everyone!!

PR #288, opened 2017-01-20
Fix CommandOrControl+R to work only when window is active

### LINKED ISSUE #287: Sqlectron takes over the Ctrl-R shortcut for the whole system
System: Ubuntu 14.04

When sqlectron is open, Ctrl-R does not work. This is specially annoying in the terminal, where bash by default has Ctrl-R assigned to reverse-search.

The problem goes away as soon as sqlectron is closed.

krolow:
Close #287

Before the Ctrl + R or Command + R was a global command, this PR makes it only work when the window has focus...

### DIFF
--- a/src/browser/app.js
+++ b/src/browser/app.js
@@ -19,14 +19,22 @@ app.on('window-all-closed', () => {
   }
 });
 
+let queryExecuteCommandRegister;
 
-// This method will be called when Electron creates a new browser window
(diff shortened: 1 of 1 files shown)


maxcnunes:
Close due other implementation without requiring global shortcuts https://github.com/sqlectron/sqlectron-gui/pull/289
Thanks @krolow!

PR #289, opened 2017-01-21
Ovewrite reload shortcut without global settings

maxcnunes:
Global shortcut should be avoided since we don't want to the shortcut work on the OS global scope.


maxcnunes:
@krolow could you check if it works for you? At least on mac it is working for me.

krolow:
it does... just that thing that I have mention when I did this solution, when you are at spalsh screen, or at the home/select database, <kbd>ctrl + r</kbd> does the reload...

krolow:
Refs to #287

PR #325, opened 2017-06-12
Use a File Path Dialogue Window to add a local Database

### LINKED ISSUE #303: Any way to select a local SQLite Database?
sqlectron-gui
1.20.2 (1.20.2.439)

OS X 10.11.6 (15G1217)

I can't find any possible way to select a local sqlite database stored as file on the local machine.
db file

The only settings possible to define are: Name, Dtabase Type and Database/Keyspace field
settings window

Maybe it is not an issue in general but a lack of UI logic to handle such common user task.

WFrsh:
When adding a new local database, the filepath can be chosen from a file dialogue instead of writing it out by hand.

#### Before

In the Database/Keyspace input, the filepath of a database had to be selected manually.

before

#### After

Now, it can be selected via a file path dialogue.

after

resolves #303

### DIFF
--- a/src/renderer/components/server-modal-form.jsx
+++ b/src/renderer/components/server-modal-form.jsx
@@ -320,7 +320,7 @@ export default class ServerModalForm extends Component {
                 placeholder="Database"
                 value={this.state.database || ''}
                 onChange={::this.handleChange} />
-              {this.isFeatureDisabled('server:password') &&
+              { this.state.client === "sqlite" &&
                 <label htmlFor="file.database" className="ui icon button btn-file">
                   <i className="file outline icon" />
                   <input
--- a/src/renderer/components/server-modal-form.jsx
+++ b/src/renderer/components/server-modal-form.jsx
@@ -314,7 +314,7 @@ export default class ServerModalForm extends Component {
           </div>
           <div className={`four wide field ${this.highlightError('database')}`}>
             <label>Database/Keyspace</label>
-            <div className="ui action input">
+            <div className={this.state.client === "sqlite" && "ui action input"}>
               <input type="text"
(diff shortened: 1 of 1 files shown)


src/renderer/components/server-modal-form.jsx
```diff
@@ -314,11 +314,22 @@ export default class ServerModalForm extends Component {
           </div>
           <div className={`four wide field ${this.highlightError('database')}`}>
             <label>Database/Keyspace</label>
-            <input type="text"
-              name="database"
-              placeholder="Database"
-              value={this.state.database || ''}
-              onChange={::this.handleChange} />
+            <div className="ui action input">
```
maxcnunes: Could you make the input file button only be visible for sqlite? I believe this button could be confuse for most users using the other database clients that does work with local files.

WFrsh:
Hi @maxcnunes ,

I changed it so that the file dialogue is only visible, when selecting the SQLite database.

src/renderer/components/server-modal-form.jsx
```diff
@@ -314,7 +314,7 @@ export default class ServerModalForm extends Component {
           </div>
           <div className={`four wide field ${this.highlightError('database')}`}>
             <label>Database/Keyspace</label>
-            <div className="ui action input">
+            <div className={this.state.client === "sqlite" && "ui action input"}>
```
WFrsh: On could also change it to 
```class={this.state.client === "sqlite" ? "ui action input" : "ui input"}```
which would keep the semantic declaration scheme. 
Don't know, if it's necessary though.

maxcnunes:
Thats @WFrsh. It is great now!

Issue #316, opened 2017-05-10
Schema doesn't filter by schema


WhyNotHugo:
I created a new connection, and set the schema to `public`:

![2017-05-09t23 30 16 818520787-03 00](https://cloud.githubusercontent.com/assets/730811/25880734/ad9560ac-3528-11e7-803e-cf48ce3f0aee.png)

But I'm still seeing tables from all schemas (including the PG ones, which are a lot):

![2017-05-09t23 30 24 202310113-03 00](https://cloud.githubusercontent.com/assets/730811/25880743/b9a8c35c-3528-11e7-8dcb-5dd1b47a35f8.png)

BTW: Good job on the GUI, I can't believe I missed something this well polished. Except for these two bugs, this is great stuff!

santaclauze:
Same issue here!

maxcnunes:
@hobarrera and @santaclauze The schema field in the server information window is only the schema used for the initial connection. But you still able to access the other available schemas. If you want to filter schemas or databases listed in the sidebar then you need to define that manually in the `~/.sqlectron.json` configuration (I have not included that in the server information window yet) with the `filter` property https://github.com/sqlectron/sqlectron-gui/blob/master/docs/app/configuration-file.md.

hobarrera:
Any plans to actually make the chosen schema the default later on? I think it makes sense to keep that.

Also, why dump the config in the `home`, instead of the standard [xdg directories](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html)?

maxcnunes:
We don't have plans for making the default schema the only available in the connection. But we could have like a checkbox in the server information window to make that possible. On selecting that checkbox it would automatically set that `filter` property.

About the xdg standard there is a issue for that already https://github.com/sqlectron/sqlectron-gui/issues/327

lordjavac:
How about grouping objects by schema in the lists. Then we can expand the schema we want to browse but still have access to the others. So, under "Tables" we can have "information_schema", "pg_catalog", "public", and any other applicable schemas. It think this would help declutter the lists while retaining the desired access.

maxcnunes:
The trick thing about the multiple database types support is that when we need to add a new feature or changing an app behavior, you have to take in consideration all the other database clients. Some times, we need to keep the UI or even internal implementation simpler so it can have a better consistency over all databases clients.
[...]

PR #356, opened 2017-10-04
Grouping database items under schema dropdowns

s-kem:
Referenced in #316 

I added in `database-list-item-metadata` a grouping on the key `schema` of all the elements in the list.

The way it's implemented, if the result of the group is `undefined`, the string `'undefined'` or an empty string (`''`) it should continue rendering as it currently does, that is, in a long, flat list.

I also modified `database-item` slightly as I borrowed code from there (the collapse arrow) and pulled that borrowed code out into a new component named `collapse-icon`, which renders the collapse icon based on a direction prop passed in.

`database-list-item-metadata`, when grouping by `schema` will render a list of `schemas` with dropdowns to the left. When clicking one, a list of all the items with that schema will then be rendered, with a `leftPadding` of 15px to differentiate that it's a sublist.

I don't have a docker instance of a database without schema support so I'd really appreciate anyone testing that for me if possible. I tested it by just making the key I was grouping on a random string, but I don't know what the app puts in the `schema` field in the data (if anything) for databases that don't have schema support.

### DIFF
--- a/src/renderer/components/collapse-icon.jsx
+++ b/src/renderer/components/collapse-icon.jsx
@@ -1,0 +1,13 @@
+import React, { PropTypes } from 'react';
+
+const CollapseIcon = ({ arrowDirection }) =>
+  (<i
+    className={`${arrowDirection} triangle icon`}
+    style={{ float: 'left', margin: '0 0.15em 0 -1em' }}
+  />);
+
+CollapseIcon.propTypes = {
+  arrowDirection: PropTypes.string.isRequired,
+};
+
+export default CollapseIcon;
--- a/src/renderer/components/database-item.jsx
+++ b/src/renderer/components/database-item.jsx
@@ -1,4 +1,5 @@
 import React, { Component, PropTypes } from 'react';
+import CollapseIcon from './collapse-icon.jsx';
 import TableSubmenu from './table-submenu.jsx';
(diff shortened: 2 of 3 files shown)


s-kem:
Good call.

I changed it to indent by 0.5 em instead of 15px and that seems to cause indentation to just cascade.

screen shot 2017-10-04 at 5 54 02 pm

BornaP:
In order to avoid column primitive type to wrap in a new row, I propose setting overflow-x to auto on whole tree list, so those inner child elements that don't fit container size, can be scrolled without native scrollbar initially showing. This way if user doesn't size sidebar to fit all the content, he can still scroll it through. Currently wrapping primitive type under a column name feels very messy.

s-kem:
@BornaP While I agree with what you're saying, the app as-is already wraps like that. Nothing in this PR is causing that.

Screen shot from the live app connecting to the same DB

screen shot 2017-10-05 at 4 51 27 pm

I'd be happy to look into it in an additional issue/PR but I gave it a quick glance and nothing stood out as immediately obvious to me on a solution.

maxcnunes:
Fair enough. We can handle the `overflow-x` mentioned by @BornaP in a separated PR.
@s-kem Thanks again :)

Issue #348, opened 2017-09-28
Resize Sidebar


seesemichaelj:
It would be nice to be able to resize the sidebar via click-and-drag on the right (or inner) border. Sometimes, the schema+table names are too long and wrap to a second line mid-character (the warp behavior is as I would expect it since these names shouldn't have spaces in them) making it slightly difficult to quickly discern the name of the table.

maxcnunes:
@seesemichaelj What version are you using? We already have that, but it stays in the top right of the sidebar. Doesn't that work for you?

seesemichaelj:
I have version 1.25.0-471 (via `dpkg`; is there a way to see this in the app? I went to Help>About which brought the website up instead). You are right though; I do see the resize functionality and it works as I'd expect it. I guess it's not intuitive to put it at the top as most interfaces have the drag throughout the whole thing. I could be happy with the current interface although if you wanted to keep it as is.

maxcnunes:
I tried to keep through the whole sidebar. But I couldn't get it working well. I would gladly accept a PR for that though :)

PR #359, opened 2017-10-08
Made resize span for database list full height, changed cursor

s-kem:
In regards to #348

Grabbed the `react-resize-handle` class and made it 100% high so it follows the entire line.

I thought I'd try a different cursor and see how you felt about it. 

To me the grab icon means "I can move this thing around the screen", not "I can resize a column" so I changed it to `col-resize` and gave it the `!important` tag to overwrite the one given by `react-resizable`

### DIFF
--- a/src/renderer/components/react-resizable.css
+++ b/src/renderer/components/react-resizable.css
@@ -47,3 +47,8 @@
   box-sizing: border-box;
   cursor: se-resize;
 }
+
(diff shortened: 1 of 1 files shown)


maxcnunes:
Yeah feels good with that cursor. And it is much better to be able to resize through the whole sidebar. Thank you!!

Issue #318, opened 2017-05-17
Remember tabs for each connection


gooddadmike:
One of my favorite features of MySQL Workbench is the way I can open a connection I haven't opened for months and all my last sql files/tabs are there. 

This would be such a nice feature to have.

s-kem:
I'm putting some thought into how to add this.

My first thought would be to just store it in the configuration file where the databases are stored, but that looks to be handled by core so my gut feeling is that's shared w/ the cli version.

Maybe save a new config file where it's just a JSON blob split up by server connection IDs?

Something like an array of objects, with `[{ serverId: selectedServerId, queries: { ...state.queries} ]]`

That'll probably get expensive writing every change to disk and state, so would probably want to debounce it, only save to disk on either tab change or query execute or sever disconnect.

Storing `state.queries` lets us just load up saved state though for whatever server was selected.

maxcnunes:
The CLI has not been maintained for a long time. So, I wouldn't care about that. At least, seems it is easier to use the existing configuration file instead of handling another one.

But at the same time, that kind of transient data could grow really fast. So for that reason would be better keep it in a separated file.

Yeah, that kind of postponing would be important to not cause performance issues.

s-kem:
@maxcnunes 

Would you rather I add on to `sqlectron.json` file or create a separate file to store this data?

If the first, then I don't need to submit any changes to `sqlectron-core` because I can just leverage the `config.save` function already exposed.

If not then I'd need to submit a PR to core to expose the `utils` file for import into the ui.

maxcnunes:
I believe we can start with the simplest solution by saving into the `sqlectron.json`. If later we find out any performance issues related to that then we can improve it by splitting it up.

maxcnunes:
The time has come, I am starting to work on this feature 🥳

gooddadmike:
Horay. I thought for some reason SQLectron was dead

maxcnunes:
After a long pause I feel refreshed to continue the project back again :)

Issue #277, opened 2016-12-16
Feature: query history


rrbarbosa:
It would be nice if we could store our query history per server. Maybe backed up by a sqlite DB, or something similar. If we agree on the basic functionality, I could help with development.

maxcnunes:
The app already has initial support for that. Every executed query is stored in memory. I had implemented that thinking in finishing this in the future. But I never got the time to finish that.
Here is the code where the query is stored after been executed: https://github.com/sqlectron/sqlectron-gui/blob/master/src/renderer/reducers/queries.js#L60
I thought it could have like a select element that would allow select any previous query executed for that tab in the current session. Is that what you were thinking?

rrbarbosa:
No, ideally I would like to keep history per server, so they are persisted over sessions. Other clients, like Postico or Sequel Pro have a button on the UI that allows you to view and search past queries, a little like a bash history. What do you think?

I think if we keep the size of the history small, ~100 queries, it should be fine to keep them in plain text.

maxcnunes:
I think that button would be even better if the user could filter by past queries for only the current tab as well. So the user have access to all queries in that session. But if he want to see only the queries for that current tab he is able to filter that. Could be a different button or just a check `show only queries of current tab`.

jonatascb:
I would like to support that feature as well. I think that would be very useful.

AgDude:
psql for postgres has similar functionality where you can use ctrl-r and start typing just like in bash. To me that would be the ideal functionality, but a UI to view and search history would be nice too.
I also would consider contributing to this.

PR #371, opened 2017-10-15
Saving tabs and queries to disk by server & automatically reloading on server selection

s-kem:
In regards to issue #318

~~Currently in progress but looking for feedback -- Mostly done, just need to thoroughly test it.~~ Finished

I'm persisting `store.queries` by `serverId` in the `sqlectron.json` file on every action that updates queries, debounced by 500ms.

When selecting a server, upon successful connection I'm retrieving the stored data and dispatching an action to put it in the store if a stored query list exists.

~~Right now I have an odd bug in that if your last selected tab was any besides the first, and you log out then back in and select tab 1, it won't pop in the text for tab 1 until you click the editor pane. I don't know why, but if I can't figure that out.~~

~~I think I'm going to just default to tab 1 on every load, because if tab 1 is the initial tab it works fine.~~ Fixed it -- See below

In this PR I also made the `selectedQuery` value get passed into the actions -- my initial plan for implementation was going to have me updating only changed parts, but decided it was easier to just store an entire snapshot of the `queries` object from the store.

During this PR I somehow introduced a bug in the `REMOVE_QUERY` reducer, at the line where the removed query was being deleted from the `queriesById` object. I fixed that by bringing in `lodash.omit` and just omitting that key instead.

My thought here is that this was failing due to mutating the underlying data, as simply spreading state to make a copy only made a shallow 1-layer deep copy, so the `delete` command was trying to directly mutate data in the store (because `queries.queriesById` is only a pointer at this point, not a copied value) -- `_.omit()` returns a brand new object without mutating, and this is generally regarded as best practice.

Here's an example of a `sqlectron.json` file after this PR w/ 2 tabs open in my second db (note: these are the docker instances, so passwords aren't secure)

```{
  "servers": [
    {
      "id": "14455516-25b2-40f4-879a-02df23f8c168",
      "name": "docker-pg",
      "client": "postgresql",
      "ssl": false,
      "host": "192.168.99.100",
      "port": 5432,
      "socketPath": null,
      "user": "user",
      "password": "14488c2275e8fd79",
      "database": "company",
      "schema": null,
      "encrypted": true
    },
    {
      "id": "fe61b1a7-f815-4c26-95c4-bb628c67a606",
      "name": "docker-mysql",
      "client": "mysql",
      "ssl": false,
      "host": "192.168.99.100",
      "port": 3306,
      "socketPath": null,
      "user": "root",
      "password": "14488c2275e8fd79",
      "database": "authentication",
      "schema": null,
      "encrypted": true
    }
  ],
  "queries": {
    "fe61b1a7-f815-4c26-95c4-bb628c67a606": {
      "lastCreatedId": 3,
      "currentQueryId": 2,
      "queryIds": [
        1,
        2
      ],
      "queriesById": {
        "1": {
          "id": 1,
          "database": "authentication",
          "name": "authentication #1",
          "filename": null,
          "isExecuting": false,
          "isDefaultSelect": false,
          "didInvalidate": true,
          "query": "query 1",
          "selectedQuery": 1,
          "queryHistory": [],
          "results": null,
          "error": null,
          "copied": false,
          "resultItemsPerPage": 100
        },
        "2": {
          "id": 2,
          "database": "authentication",
          "name": "authentication #2",
          "filename": null,
          "isExecuting": false,
          "isDefaultSelect": false,
          "didInvalidate": true,
          "query": "query 2",
          "selectedQuery": 2,
          "queryHistory": [],
          "results": null,
          "error": null,
          "copied": false,
          "resultItemsPerPage": 100
        }
      },
      "resultItemsPerPage": 100,
      "enabledAutoComplete": true,
      "enabledLiveAutoComplete": true
    }
  }
}```

### DIFF
--- a/src/renderer/actions/tables.js
+++ b/src/renderer/actions/tables.js
@@ -1,4 +1,4 @@
-import { getCurrentDBConn } from './connections';
+import { getDBConnByName } from './connections';
 
 
 export const FETCH_TABLES_REQUEST = 'FETCH_TABLES_REQUEST';
@@ -31,10 +31,10 @@ function shouldFetchTables (state, database) {
 
 
 function fetchTables (database, filter) {
-  return async (dispatch, getState) => {
+  return async (dispatch) => {
     dispatch({ type: FETCH_TABLES_REQUEST, database });
     try {
-      const dbConn = getCurrentDBConn(getState());
+      const dbConn = getDBConnByName(database);
       const tables = await dbConn.listTables(filter);
       dispatch({ type: FETCH_TABLES_SUCCESS, database, tables });
     } catch (error) {
--- a/package.json
+++ b/package.json
@@ -126,6 +126,8 @@
     "lodash.debounce": "^4.0.8",
     "lodash.groupby": "^4.6.0",
     "lodash.isplainobject": "^4.0.6",
+    "lodash.mapvalues": "^4.6.0",
+    "lodash.omit": "^4.5.0",
     "lodash.set": "^4.3.2",
     "lodash.template": "^4.4.0",
     "lodash.trim": "^4.5.1",
--- a/package.json
+++ b/package.json
@@ -132,6 +132,7 @@
     "lodash.template": "^4.4.0",
     "lodash.trim": "^4.5.1",
     "lodash.union": "^4.6.0",
+    "lodash.uniq": "^4.5.0",
     "minimist": "^1.2.0",
     "node-sass": "^3.4.2",
     "react": "^15.0.1",
--- a/src/renderer/actions/connections.js
+++ b/src/renderer/actions/connections.js
@@ -59,8 +59,13 @@ export function restoreStoredQueries(id) {
       const databaseNames = uniq(
         Object.values(storedQueryState.queriesById).map(({ database }) => database)
       );
-      databaseNames.forEach(db => dispatch(connect(id, db)));
-
+      databaseNames.forEach((db, idx) => {
+        if (idx === 0) {
+          dispatch(connect(id, db, false, undefined, true));
+        } else {
+          dispatch(connect(id, db));
+        }
+      });
       dispatch({ type: SET_STORED_QUERIES, storedQueryState });
     } else {
       dispatch(connect(id));
@@ -68,7 +73,8 @@ export function restoreStoredQueries(id) {
   };
 }
 
-export function connect (id, databaseName, reconnecting = false, sshPassphrase) {
+export function connect (id, databaseName, reconnecting = false,
+  sshPassphrase, isServerConnection = false) {
   return async (dispatch, getState) => {
     let server;
     let dbConn;
@@ -100,7 +106,7 @@ export function connect (id, databaseName, reconnecting = false, sshPassphrase) 
         server,
         database,
         reconnecting,
-        isServerConnection: !databaseName,
+        isServerConnection: isServerConnection || !databaseName,
       });
 
       if (!serverSession) {
--- a/src/renderer/actions/connections.js
+++ b/src/renderer/actions/connections.js
@@ -1,5 +1,6 @@
 import { sqlectron } from '../../browser/remote';
-import getCurrentServerId from '../utils/getCurrentServerId';
+import { SET_STORED_QUERIES } from './queries';
+import uniq from 'lodash.uniq';
 
 export const CLOSE_CONNECTION = 'CLOSE_CONNECTION';
 export const CONNECTION_REQUEST = 'CONNECTION_REQUEST';
@@ -38,8 +39,7 @@ export function getDBConnByName(database) {
   return dbConn;
 }
 
-const getStoredQueryState = state => {
-  const currentServerId = getCurrentServerId(state);
+const getStoredQueryState = (state, currentServerId) => {
   try {
     const storedQueryState = state.config.data.queries[currentServerId];
     if (storedQueryState && Object.keys(storedQueryState).length > 0) {
@@ -51,6 +51,22 @@ const getStoredQueryState = state => {
     return null;
   }
 };
+
+export function restoreStoredQueries(id) {
+  return async (dispatch, getState) => {
+    const storedQueryState = await getStoredQueryState(getState(), id);
+    if (storedQueryState) {
+      const databaseNames = uniq(
+        Object.values(storedQueryState.queriesById).map(({ database }) => database)
+      );
+      databaseNames.forEach(db => dispatch(connect(id, db)));
+
+      dispatch({ type: SET_STORED_QUERIES, storedQueryState });
+    } else {
+      dispatch(connect(id));
+    }
+  };
+}
 
 export function connect (id, databaseName, reconnecting = false, sshPassphrase) {
   return async (dispatch, getState) => {
@@ -110,15 +126,7 @@ export function connect (id, databaseName, reconnecting = false, sshPassphrase) 
       dbConn = serverSession.createConnection(database);
       await dbConn.connect();
 
-      const connectionSuccessAction =
-        { type: CONNECTION_SUCCESS, server, database, config, reconnecting };
-      const storedQueryState = await getStoredQueryState(getState());
-
-      if (storedQueryState) {
-        dispatch({ ...connectionSuccessAction, storedQueryState });
-      } else {
-        dispatch(connectionSuccessAction);
-      }
+      dispatch({ type: CONNECTION_SUCCESS, server, database, config, reconnecting });
     } catch (error) {
       dispatch({ type: CONNECTION_FAILURE, server, database, error });
       if (dbConn) {
--- a/src/renderer/actions/connections.js
+++ b/src/renderer/actions/connections.js
@@ -1,5 +1,5 @@
 import { sqlectron } from '../../browser/remote';
-
+import getCurrentServerId from '../utils/getCurrentServerId';
 
 export const CLOSE_CONNECTION = 'CLOSE_CONNECTION';
 export const CONNECTION_REQUEST = 'CONNECTION_REQUEST';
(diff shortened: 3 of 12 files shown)

### CODE REVIEW

File: src/renderer/components/query.jsx [hunk 1]
```diff
@@ -68,11 +68,6 @@ export default class Query extends Component {
   }
 
   componentDidMount() {
-    this.refs.queryBoxTextarea.editor.on(
```
s-kem: I'm not sure what the purpose of this was, but by removing it I removed a duplicate dispatch of the `UPDATE_QUERY` action on every key press, and solved an issue where I wasn't loading the correct `query.query` into the `AceEditor` pane per tab.

s-kem:
Fixed that bug I was seeing w/ the tabs -- connection success spawns a new tab, but then it was being overridden right after by setting the cached state and a re-render just wasn't happening -- so I just tied in to the `CONNECTION_SUCCESS` action for setting the initial state -- it's either a call to `addNewQuery` or it's the result of spreading initial state w/ the stored values.

s-kem:
How do you feel about storing the result of the queries to disk?

My thought was that would cause the JSON to get potentially way too massive and for someone who uses a lot of tabs would cause pretty long load times when starting the app, especially if they had a couple `SELECT *` queries open on large databases.

maxcnunes:
Yeah, I think that could lead to performance issues pretty easy. I don't think is necessary to save the results. With the restored queries, the user should be able to execute it again with no problem.

maxcnunes:
@s-kem This week I'm quite busy. I will review/merge it up to Friday though.

s-kem:
@maxcnunes No problem! Thanks for letting me know.

s-kem:
I did break it but it wasn't that `this.refs.queryBoxTextArea.editor.on()` function that did it.

I was confused as to what `selectedQuery` was supposed to be within each `queryById` object, I thought it was supposed to be the currently selected query tab, and didn't notice it's inside each query object.

I thought it was just not being used to in all cases to pass back the current tab ID, so I made changes to make sure it did. The `cmd+enter` keyboard shortcut was relying on a different value that was now the query ID.

I should have undid those changes to start with because they weren't even necessary for the implementation I ended up going with, so I reverted.

I also caught that I wasn't correctly clearing errors/previous results, so I brought in `mapValues` from lodash in order to map over each `queriesById` object and set the `result` to an empty array and `error` to null.

maxcnunes:
Hey @s-kem, I took some time to review it today. The code looks good. But I have found few bugs 😢 

**Related to queries stored state:**
*  The "Select rows" from the sidebar context menu does not trigger the action to save the query. I have not tested the other context menu options. But believe they probably don't work too.
* The restoring state from a database connection that is not the initial one fails. 

**Related to the "execute selected text"**
Seems it is not working anymore. Looks like it is always executing the whole query box content. Here are GIFs showing the before and after:
Before your changes
before
After your changes
after

s-kem:
Ooohhh, man, I feel so dumb for that. I get why that dispatch was happening now. Also, I didn't even know that was a feature of this app, so that's awesome!

In hindsight I would've introduced some unwanted behaviour in saving the state had I not done that. I'm also clearing `selectedQuery` from each query prior to writing to disk.

Ive reverted removing that code... It wasn't causing an error anymore in my implementation...

in fact I just squashed all the commits down into one so the history doesn't get messy. Should be fine since nothing else was based on this work yet.

Thanks for your patience in helping me get through it.

Is there any other data you can think of we shouldn't be saving to disk? I was considering clearing `queryHistory` so it doesn't balloon up over time... or at least pruning it to say the last 10.

maxcnunes:
No worries, you have been doing a great work so far :)

By this quote below, did you mean it will not include the context menu queries?

> In hindsight I would've introduced some unwanted behaviour in saving the state had I not done that.  

I believe on MySQL Workbench it restores all tabs, even the empty ones. I think that is a more expected behaviour at least.

`queryHistory` was supposed to be used for this feature https://github.com/sqlectron/sqlectron-gui/issues/277. But that still has not been implemented yet. In that case, I believe you could just strip that out of the file system stored data.

Sorry, when I wrote down the description for this bug below. I knew it wasn't clear, but I was so lazy that I was hoping you could understand it 😆 

> The restoring state from a database connection that is not the initial one fails

Let's say I have a server with 2 databases: A and B; and have configured in the server connection to A be the initial database.

1. I open the connection to that server which will automatically open a tab to database A.
1. I select the database B in the sidebar and write down a query into the tab to database B.
1. I close the whole server connection.
1. I connect again to that server. It will start to fail the connection loading.
1. I force to close the app.
1. I open the app again and change that server connection setting to use database B as the initial one.
1. I connect again to that server and now the app works.

I guess you just need to open the database connection to all databases related to the restored tabs.Would be even better if we could restore the tabs in a lazy loading approach. Which would only open the database connection if the user selects a tab related to that database. But that makes the implementation more complicated. So, I'm ok with the first approach.

s-kem:
> By this quote below, did you mean it will not include the context menu queries?
> 
>> In hindsight I would've introduced some unwanted behaviour in saving the state had I not done that.
> I believe on MySQL Workbench it restores all tabs, even the empty ones. I think that is a more expected behaviour at least.

I should've been clearer there, sorry.

Now that I actually know what `selectedQuery` is (the hilighted text), I noticed if I hilighted a statement, hit `cmd+enter` to execute it, quit the app and re-opened, `selectedQuery` was still the statement I had previously hilighted, but I wasn't restoring the hilighting.

So if I had two statements:

```SELECT * FROM roles;
SELECT * FROM users;```

and I hilighted the first and ran it, then I only execute the first. When re-opening the app, nothing is hilighted but when I execute it would only be executing the first.

So I need to do one of two things:

1. Figure out how to restore the hilighting in the UI
2. Clear selectedQuery so executing will just execute the whole block and let the user hilight a specific query again

I opted for 2 since it felt easier.


On the other bug,  I fully understand it now, it's outside my normal workflow in the app so I forgot all about testing that.

Fixing that was a big challenge.

I re-set `ConnActions.connect` back to what it was before... I initially tied into it since itw as the first action that runs.

Instead I added a new action, `ConnActions.restoreStoredQueries` which takes a server id as a param.

I updated `QueryBrowser` to dispatch that on `componentWillMount` instead of the `connect` action. This new action grabs the stored state and loops through each database tabs were open for and dispatches `connect` for each of them.

In the event that there's no stored state it just dispatches `connect` for the passed in id, which will then connect to the default database.

To get around the `componentWillReceiveProps` in `QueryBrowser` navigating me back to `/` when I restored query, I had to have the reducer for `connection` set `connecting` to `true` when dispatching `SET_STORED_QUERIES` -- this seemed the simplest way to ensure it knew connections were pending, because `SET_STORED_QUERIES` would dispatch before any `CONNECTION_REQUEST` actions due to the async calls in the `connect` function.

**Comment by maxcnunes** [comment]
Related to the "select" issue. I agree the second approach seems a better option.

Ok, it is looking better. But I have found other 2 bugs 😢 

**1. Is not resetting the lastCreatedId prop from reducer/queries when there are no tabs to be restored**

Makes sense to keep the `lastCreatedId` when there are tabs to be restored. But it is weird to keep that value if there are no tabs to restore. 

To reproduce the bug:
1. connect to a server
1. open 3 tabs and close them all
1. disconnect from that server
1. connect to that server again
It will start from query Id 4 instead of from 1.

**2. Is loading the same database info (tables, views, etc.) from the selected tab to all databases**

To reproduce the bug:
1. connect to a server:
1. open a tab for database A and type down a query into it
1. open a tab for database B and type down a query into it
1. disconnect from that server
1. connect to that server again
1. take a look in the sidebar, both databases will have the same data

**Comment by s-kem** [comment]
I added a commit fixing the first -- Any reason you think that change might break something else?

I seem to be calling back the correct tables when opening a brand new connection (as the app currently does) as well as calling back the correct tables on all databases I'm opening when restoring previous state.

I think the rational for the original line was the `currentQueryId` when that function would fire was always the newly spawned tab, but when restoring I'd need the tables for both.

**Comment by tengis** [comment]
@maxcnunes Can you please release this feature? 🙏

**Comment by maxcnunes** [comment]
I wasn't able to test it again. I got very busy at work these last days. I will test it as soon as I get some free time.

**Comment by maxcnunes** [comment]
Sorry @s-kem, I have been traveling and I will not have time to review it for the next month. Since you have been helping a lot in the project. I will make you a contributor 🎉 
Let's just wait like a week if someone else can review it. Otherwise, you can go ahead and merge it.

**Comment by maxcnunes** [comment]
Hey @BornaP, would be awesome if you could take a look on this one too 😅

**Comment by BornaP** [comment]
Hehe, this one has a looong history :) I'll take a look as soon as possible.

Congrats @s-kem for joining contributors crew ;)

File: src/renderer/actions/tables.js [hunk 2]
```diff
@@ -31,10 +31,10 @@ function shouldFetchTables (state, database) {
 
 
 function fetchTables (database, filter) {
```
BornaP: @s-kem I think you'll need to do the same fix for views, procedures etc.

s-kem: You're probably right. I added a commit for that. I'll test it tomorrow on my production databases which have some views and triggers across different databases. My personal dev machine just has the docker instances. I should probably add some views and functions to those so I can catch things like this.

BornaP: Have you tested it on production databases yet? :)

I'll test it out on my local dev instance.

BornaP: Looks alright to me. Please confirm if you have tested on production databases :)

s-kem:
Hey @BornaP , sorry, been a crazy week.

I did test it out on my work databases and it appears to be loading the data as needed, but most of them didn't have anything to fill out all the categories anyway.

Let me know if you can think of anything else this needs.

maxcnunes:
@s-kem sorry we couldn't merge your PR. I plan to work on this feature now, but I will be following a slight different approach so I will need to close your PR. I appreciate all the work you have put into this though ❤️

PR #372, opened 2017-10-15
fixed the redux logger to include sub-items under action logs

s-kem:
Redux logger seems to need the initial action log to be a `console.log` and not a `console.debug`.

By changing it to a `console.log` you can get the dropdowns under the actions that show the before and after of the store, as well as the value of the action.

Before (you can see 3 of the items are expanded, but nothing appears):

screen shot 2017-10-15 at 3 01 19 pm

After:
screen shot 2017-10-15 at 3 04 48 pm


maxcnunes:
Weird, the logs already appear for me without this change. But I guess there is no problem adding that handle as well.

s-kem:
I wonder if it's an electron thing. I'm running on OSX High Sierra but I can check the previous code on my Windows machine.

Curious why it wasn't working, but since I started contributing I've never been able to get actions to drop down so I could see the prev/next state or the action object.

Issue #298, opened 2017-02-17
ability to wrap content in SQL query editor


Ramanathan21:
It would be helpful, if we can toggle "wrap contents" in the SQL Query editor. Currently now it's bit hard to keep scrolling horizontly.

Trevormichael:
I got this one

PR #373, opened 2017-10-18
Added checkbox to toggle wrapping in query editor (#298)

Trevormichael:
See the screenshot below.

I've added a toggle button that enables / disables wrapping in the query editor as per #298 so that it is no longer necessary to keep scrolling horizontally. 

I can gladly move this toggle elsewhere in the ui or put it a menu somewhere - I wasn't too sure about where it was supposed to be. 

Also, I did take into account the editor box resizing and tested to make ensure that didn't mess up the placement. 

screenshot from 2017-10-18 00-15-31

### DIFF
--- a/src/renderer/components/query.jsx
+++ b/src/renderer/components/query.jsx
@@ -6,6 +6,7 @@ import 'brace/mode/sql';
 import 'brace/theme/github';
 import 'brace/ext/language_tools';
 import 'brace/ext/searchbox';
+import CheckBox from './checkbox.jsx';
 import QueryResult from './query-result.jsx';
 import ServerDBClientInfoModal from './server-db-client-info-modal.jsx';
 
@@ -64,7 +65,9 @@ export default class Query extends Component {
 
   constructor(props, context) {
     super(props, context);
-    this.state = {};
+    this.state = {
+      wrapEnabled: false
+    };
   }
 
   componentDidMount() {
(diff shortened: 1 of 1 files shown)


BornaP:
Can you explain the PR bit more. What toggling the checkbox should do?

Also, for performance issues, you should probably make handler methods and use `::this.[methodName]` syntax like the rest of the code does. Your current code makes two new functions in memory on each render cycle.

Trevormichael:
@BornaP Thank you for pointing that out! I just submitted some changes as per your suggestion. 

And I apologize, I thought it was clear but I've updated my initial  comment to indicate what the toggle button is actually for.

src/renderer/components/query.jsx
```diff
@@ -275,17 +286,26 @@ export default class Query extends Component {
               mode="sql"
               theme="github"
               name={this.props.editorName}
-              height="100%"
+              height="calc(100% - 15px)"
```
maxcnunes: wow this trick is new to me 😄

BornaP: Pretty long in the standard https://caniuse.com/#search=calc ;)

BornaP:
Seems fine to me. I'll merge it. Thank you @Trevormichael

maxcnunes:
@Trevormichael I tested it and looks like it is not working:



hey @BornaP, did it work with you?

BornaP:
Argh, seemed so simplistic, I didn't bothered to test manually. Sorry, my bad 😲

maxcnunes:
No problem, I do that sometimes (pretty often) too 😆

Trevormichael:
@maxcnunes In that test you posted, none of the text is overflowing horizontally so toggling wrapping is not going to do anything

maxcnunes:
Alright. That is working. Thanks!!

PR #378, opened 2017-10-20
Improve pronunciation

nathansimpson5:
added what I think is a better pronunciation guide

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,5 +20,4 @@
 
 #### How to pronounce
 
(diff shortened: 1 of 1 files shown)


maxcnunes:
That really sounds better. Could you please replace that with:
```
It is pronounced "sequel-eck-tron" - https://translate.google.com/?source=osdd#en/en/sequel-eck-tron
```

nathansimpson5:
Made the change you suggested. Sorry, but I'm new to this, do I need to make a new pull request, or is the commit I made ok?

maxcnunes:
That commit is ok. Thanks!!

Issue #364, opened 2017-10-14
Block drag n Drop feature. 


SumantaBee:
Block the drag n drop feature into the chrome window. I dragged a notepad file into the window and it started showing the notepad content. No way to get back to the app html again.

synle:
Can you elaborate more on what is the expected behavior?

SumantaBee:
If you running an Electron app, and you drag and drop any file from your desktop into the electron windows the windows will load the dropped file instantly and there will be noe 'Go Back' option. Think about it.

Get Outlook for Android<https://aka.ms/ghei36>

________________________________

maxcnunes:
@SumantaBee The drag and drop have been disabled by #388. So, we shouldn't have the issue with the window getting stuck without a go back option.
@synle He (@SumantaBee) want to be able to open sql files on Sqlectron by dragging and dropping a file from the filesystem into the app.

SumantaBee:
correct.

Get Outlook for Android<https://aka.ms/ghei36>

________________________________

PR #388, opened 2017-10-30
Prevent drag and drop causing redirect

haivp3010:
Fix for issue #364

### DIFF
--- a/src/renderer/containers/app.jsx
+++ b/src/renderer/containers/app.jsx
@@ -25,6 +25,9 @@ class AppContainer extends Component {
 
   componentDidMount() {
     this.props.dispatch(ConfigActions.loadConfig());
+    // Prevent drag and drop causing redirect
+    document.addEventListener('dragover', this.handleDragOver, false);
+    document.addEventListener('drop', this.handleDrop, false);
   }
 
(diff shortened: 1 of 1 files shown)


s-kem:
Good catch on being safe and removing the listeners on unmount.

IMO there's no need for two functions that do the same thing, nor is there a reason for that function to live within the React class. There's no reason to use `this.someFunction()` if `someFunction` doesn't actually need to know about `this`

`const preventDefault = e => e.preventDefault();` at the top of the file, outside the scope of the class and

``` jsx
  componentDidMount() {
    this.props.dispatch(ConfigActions.loadConfig());
    document.addEventListener('dragover', preventDefault, false);
    document.addEventListener('drop', preventDefault, false);
  }

  componentWillUnmount() {
    document.removeEventListener('dragover', preventDefault, false);
    document.removeEventListener('drop', preventDefault, false);
  }
```

is cleaner imo

haivp3010:
Oh yes, I missed that. Thanks, I will change that now.

maxcnunes:
Thanks for the CR @s-kem. Did you check it is working? I will merge it if you approve that.

PR #395, opened 2017-11-07
Re-rendering QueryResult when sidebar width changes

s-kem:
Before this PR, re-sizing the sidebar doesn't cause QueryResult to re-render:

before

After this, it'll re-render when we stop dragging:

after

I'm experimenting with ways to re-render as we drag but changing `query-browser`'s `onResizeStop` to `onResize` was really laggy. I tried debouncing it but it just didn't fire at that point until I stopped dragging.

Open to ideas, but in the mean time this small change is better than leaving it as is IMO.


BornaP:
Great, thanks @s-kem ! I agree, no need for throttle/debounce.

Issue #384, opened 2017-10-25
Background color change ?


gabino:
how to change the background color, as it tires the view

maxcnunes:
Nowadays is not possible to change the background color. I mean, if you really want that, you could do that by opening the inspector mode and changing the CSS background color. Would be nice though to support different themes.

synle:
I added a few improvements to make the dark theme more pleasant. Below is the PR.
https://github.com/sqlectron/sqlectron-gui/pull/408

PR #401, opened 2017-12-03
Dark theme for body element (#384)

alexmylonas:
Suggestion for quick dark theme option.
Refactor for css and/or body element maybe is more suitable as a core change.

### DIFF
--- a/src/renderer/components/settings-modal-form.jsx
+++ b/src/renderer/components/settings-modal-form.jsx
@@ -65,6 +65,7 @@ export default class SettingsModalForm extends Component {
       limitQueryDefaultSelectTop: parseInt(state.limitQueryDefaultSelectTop, 10) || 100,
       enabledAutoComplete: state.enabledAutoComplete || false,
       enabledLiveAutoComplete: state.enabledLiveAutoComplete || false,
+      enabledDarkTheme: state.enabledDarkTheme || false,
     };
     if (!this.state.log) { return config; }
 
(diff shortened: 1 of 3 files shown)


maxcnunes:
@alexmylonas thanks for this PR. But there is an issue need to be fixed before I can accept it. After changing the theme it is not loading properly the main window:

maxcnunes:
Sorry, I just noticed this was occuring on master and your PR #400 has fixed it. Thanks 👍

maxcnunes:
That theme is not looking so good. But it is a good start. I will merge it and see if I can make a bit prettier. Thanks again!!

alexmylonas:
Hi there! Yes, i agree it can be prettier if you have any suggestion i can try to improve it in the coming days!

synle:
I looked at the PR, looks like you only change a few elements in CSS, might be a better choice if we have an app-wide black theme.

synle:
After some works, I got some good successes, but we might need to adjust more custom components.

screen shot 2017-12-21 at 1 44 53 pm

screen shot 2017-12-21 at 1 46 44 pm


screen shot 2017-12-21 at 1 45 19 pm


screen shot 2017-12-21 at 1 46 47 pm

synle:
Guys, I took a stab at it and here is my first PR for dark theme improvements.
https://github.com/sqlectron/sqlectron-gui/pull/408

@maxcnunes @alexmylonas

Issue #409, opened 2017-12-22
[Bug] Edit Connection doesn't work for RECENTLY DUPLICATED CONNECTIONS / NEW CONNECTIONS  (PR attached)


synle:
**Edit Button DOES NOT for RECENTLY DUPLICATED / NEW CONNECTIONS**

Steps to repro
1. duplicate an existing connection / or create a brand new connection
2. try to edit it (notice it's not working)...

Connect button works, but the edit button doesn't work.

I'll check to see if I can send in a PR.

synle:
Look like it could be a logic issue in the duplicate code.

After duplicate, the number of server is still showing up as 3 (instead of 4) like in this example...
![screen shot 2017-12-22 at 6 56 10 am](https://user-images.githubusercontent.com/3792401/34302360-4c0f6ac0-e6e5-11e7-9912-a5f7a39aee76.png)

synle:
After some diggings, I found that the routine we use to find the servers is not correct.

We looked at at config (for some reasons it's out of sync)
```
const server = getState().config.data.servers.find(srv => srv.id === id);
```

Whereas we should use this value instead...
```
const server = getState().servers.items.find(srv => srv.id === id);
```

synle:
Created a PR for this, PR link is here https://github.com/sqlectron/sqlectron-gui/pull/410

synle:
Look like this issue also appears in RECENTLY CREATED CONNECTIONS.

synle:
The fix should be pushed already in #410 .

What I did is to reload the config after save and after duplication

synle:
PR merged, closing out

PR #410, opened 2017-12-22
Fix Edit Button for RECENTLY DUPLICATED / NEW CONNECTIONS

synle:
Refer to this issue for more details...
https://github.com/sqlectron/sqlectron-gui/issues/409

Below is a short description
Edit and Other commands don't work for RECENTLY DUPLICATED CONNECTIONS

1. duplicate an existing connection
2. try to edit it (notice it's not working)...

Connect button works, but the edit button doesn't work.

I'll check to see if I can send in a PR.

### DIFF
--- a/src/renderer/actions/servers.js
+++ b/src/renderer/actions/servers.js
@@ -41,7 +41,7 @@ export function finishEditing() {
 }
 
 
-export function saveServer ({ server, id }) {
+export function saveServer ({ server, id }, onAfterSave) {
   return async (dispatch, getState) => {
     dispatch({ type: SAVE_SERVER_REQUEST, server });
     try {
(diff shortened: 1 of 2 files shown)


synle:
Hold up, looks like there are some encryptions / decryptions not being set properly in `servers.items` compared with `config.data.servers`

screen shot 2017-12-22 at 1 47 46 pm

synle:
Nvm, I think data within `servers.items` might have been decrypted already. This should work.

synle:
Don't merge this yet. I think I found a bigger issue with our encryption and decryption logics when DUPLICATING, and UPDATING servers.

They don't work. Try to duplicate ABC, to ABC1, then ABC1 to ABC2, the encrypted passwords are not properly copied over, the same thing is true for Trying to edit / update the same connections multiple times.

synle:
Look like the logic to encrypt passwords is encapsulated in the core library `sqlectron-core` and not exported.

So we might not be able to use `servers.items` directly.

synle:
Okay guys, I finally ended up with the RELOAD after SAVING and after DUPLICATING path.

I think it's a safer path and solution to this problem.

maxcnunes:
Good catch @synle. Thanks again 👏 👏 👏

Issue #413, opened 2017-12-22
[Feature Request] - Use Slider for Zoom Scale instead of Number (PR attached)


synle:
Maybe use Slider will make it easier to adjust.

**New Design**
![screen shot 2017-12-22 at 10 34 06 am](https://user-images.githubusercontent.com/3792401/34308796-d5da517a-e703-11e7-87a7-cd087baf6d72.png)

**Old Design**
![screen shot 2017-12-22 at 10 35 06 am](https://user-images.githubusercontent.com/3792401/34308786-cd4cbfac-e703-11e7-82f3-7f663c586b9f.png)

Or simply use HTML5 range will also make the UI much easier to see.
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/range

synle:
Pushed the PR for this, it can be found here:
https://github.com/sqlectron/sqlectron-gui/pull/414

PR #414, opened 2017-12-22
Change Zoom Factor Control to slider

synle:
Fixes https://github.com/sqlectron/sqlectron-gui/issues/413

I decided to use HTML5 range control (no dependency) instead of third party Semantic UI library (lots of dependencies that needs to be imported for us to use it).

**Benefits:**
1. No need to do extra validation on the form
2. Look nicer and easier to use from users' perspective.

**Screenshot:**
screen shot 2017-12-22 at 10 47 26 am


synle:
Fixed the lint error and pushed...

maxcnunes:
Looking good. Thanks :)

PR #415, opened 2017-12-22
Remove redundant text in setting page...

synle:
Some of the labels in Settings page are redundant, no need to add them...

screen shot 2017-12-22 at 10 56 22 am


maxcnunes:
Good call. Thanks @synle !!!

PR #424, opened 2018-01-22
fix: upgrade jquery to 3.3 to avoid security issues

eMerzh:
Let's just upgrade jquery to avoid security issue/warning :)

everything seems to still works after this.


eMerzh:
Shall I merge?

PR #450, opened 2018-09-04
Upgrade electron-builder

armarti:
Older versions of electron-builder get stuck on the "building DMG" step when building on macOD High Sierra, e.g. https://github.com/electron-userland/electron-builder/issues/2320. Later versions don't have this problem.


maxcnunes:
Thanks

PR #451, opened 2018-09-17
Feature/ui improvements

barneydmedia:
Making a PR just in case anyone wants to take a look before the changes get merged. I'm working on closing a few issues and adding some features to the UI.

### DIFF
--- a/src/renderer/components/query-result-table.jsx
+++ b/src/renderer/components/query-result-table.jsx
@@ -336,7 +336,8 @@ export default class QueryResultTable extends Component {
    * It gives a better UX since the column adapts better to the table width.
    */
   resolveCellWidth(fieldName, fields, rows) {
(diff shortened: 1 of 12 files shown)


src/renderer/components/query-result-table.jsx
```diff
@@ -336,7 +336,8 @@ export default class QueryResultTable extends Component {
    * It gives a better UX since the column adapts better to the table width.
    */
   resolveCellWidth(fieldName, fields, rows) {
-    const font = '14px \'Lato\', \'Helvetica Neue\', Arial, Helvetica, sans-serif';
+    const { customFont } = this.props.config.data.customFont;
```
rathboma: @maxcnunes  This line here actually breaks the GUI on a fresh checkout as `customFont` doesn't exist.

maxcnunes: Good catch, are able to fix it? 🙏

PR #470, opened 2019-05-03
Fixes #469 - query results error out if no custom font is set

### LINKED ISSUE #469: Running a query results in the error "Cannot match against 'undefined' or 'null'."
When checking out the project fresh on Master using Node 8 and trying to run a select query on either Postgres or MySQL causes this error in the console:

`"Cannot match against 'undefined' or 'null'."`

Screen Shot 2019-05-02 at 8 36 31 PM

rathboma:
- if no custom font has been set from the settings modal you are unable to see query results
- this fixes the issue by defaulting the custom font to the default font

Fixes #469

### DIFF
--- a/src/renderer/components/query-result-table.jsx
+++ b/src/renderer/components/query-result-table.jsx
@@ -337,7 +337,7 @@ export default class QueryResultTable extends Component {
    */
   resolveCellWidth(fieldName, fields, rows) {
     const customFont = this.props.config.data.customFont || 'Helvetica Neue';
-    const font = `14px '${customFont}' 'Helvetica Neue', Arial, Helvetica, sans-serif`;
+    const font = `14px '${customFont}', 'Helvetica Neue', Arial, Helvetica, sans-serif`;
     const numRowsToFindAverage = rows.length > 30 ? 30 : rows.length;
     const maxWidth = 220;
     const headerWidth = this.getTextWidth(fieldName, `bold ${font}`);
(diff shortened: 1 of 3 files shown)


rathboma:
Not sure what this build failure is about -- seems maybe related to a library upgrade?

rathboma:
Ah, looks like the Travis build is running on node 6 - `- nvm install 6` in the config, but there are some async functions in the code.

src/renderer/components/query-result-table.jsx
```diff
@@ -336,13 +336,11 @@ export default class QueryResultTable extends Component {
    * It gives a better UX since the column adapts better to the table width.
    */
   resolveCellWidth(fieldName, fields, rows) {
-    const { customFont } = this.props.config.data.customFont;
-    const font = `14px '${customFont}', 'Helvetica Neue', Arial, Helvetica, sans-serif`;
+    const customFont = this.props.config.data.customFont || 'Helvetica Neue';
+    const font = `14px '${customFont}' 'Helvetica Neue', Arial, Helvetica, sans-serif`;
```
maxcnunes: You removed the `,` from the customFont. I think the comma is required there.

rathboma: yes you are correct. I will fix that.

MasterOdin:
Manually merged in df7005281cfcd9a50eb625b9ac0010175d650cec

PR #488, opened 2020-05-26
update compile time dependencies

MasterOdin:
This updates "compile time" dependencies, focusing principally on babel, eslint, and webpack. Upgrading babel and its plugins requires updating to webpack, and then babel-eslint needs upgrading to run on v7, which then requires upgrading eslint to v4 (and then all of its related plugins). Upgrading these however includes three large-ish BC breaks, namely Babel 6 -> 7 and webpack 1 -> 2/3, and then 2/3 -> 4. While it would have probably been nice to have upgraded them one at a time, each of these dependencies are so intertwined that it's easier to upgrade all at once and be done with it.

As part of the upgrade of eslint, I did fix some of the new lint errors, but I added a number of ignore rules that can be solved over time.


eMerzh:
Yey 👍 
This looks great :)

Issue #566, opened 2021-02-13
Add "save as" method for saving queries to new file


MasterOdin:
After opening or saving a query, that tab now becomes strictly linked with the file. However, it would be good to support the ability to do a "Save as..." operation to allow the user to associate the tab with a different file.

lucaxchaves:
I'll take this one.

MasterOdin:
Closed by #572

PR #572, opened 2021-02-22
Add "save as" to save queries into a new file

lucaxchaves:
Added the "save as" feature to save the current query into a new file.

The shortcuts are:

Windows: Ctrl+Shift+S
Linux: Ctrl+Shift+S
Mac OS: Cmd+Shift+S

Related issue: #566

### DIFF
--- a/src/renderer/actions/queries.js
+++ b/src/renderer/actions/queries.js
@@ -176,7 +176,7 @@ export function saveQuery (isSaveAs) {
         { name: 'All Files', extensions: ['*'] },
       ];
 
-      let filename = getFileName(currentQuery, isSaveAs, filters);
+      let filename = await getFileName(currentQuery, isSaveAs, filters);
       if (path.extname(filename) !== '.sql') {
         filename += '.sql';
       }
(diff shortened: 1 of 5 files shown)


src/renderer/actions/queries.js
```diff
@@ -169,7 +176,7 @@ export function saveQuery () {
         { name: 'All Files', extensions: ['*'] },
       ];
 
-      let filename = (currentQuery.filename || await fileHandler.showSaveDialog(filters));
+      let filename = getFileName(currentQuery, isSaveAs, filters);
```
MasterOdin, line 179: Need to add an `await` here as `getFileName` returns `Promise<string>`. Without it, get and error where `filename` is a promise object and passing that to `fileHandler.saveFile` breaks.

```suggestion
      let filename = await getFileName(currentQuery, isSaveAs, filters);
```

MasterOdin (changes requested):
Thanks for the change. With the suggested change, it works perfectly.

src/renderer/actions/queries.js
lucaxchaves: Good catch!!

MasterOdin (approved):

MasterOdin:
Thanks for contributing!

PR #574, opened 2021-02-24
Configure prettier and husky

maxcnunes:
Enforce the whole codebase to follow the same format code style.

All files were updated based on prettier suggestions. I have also included husky to run linter locally before committing new changes.


MasterOdin:
Curious why dropping `eslint-config-airbnb` as it should be compatible with prettier?

maxcnunes:
@MasterOdin I initially used airbnb extension because at that time there wasn't an basic good configuration provided by eslint. Since eslint provides one now `eslint:recommended` it should be fine dropping that in favor of eslint's rules.

MasterOdin (approved):

PR #575, opened 2021-02-26
Add initial e2e tests

maxcnunes:
Add initial e2e tests. Although, for more advanced testing coverage we probably will need to upgrade Electron so it gets a recent Chromium version which Playwright depends on, for instance `click` wasn't working because of that, I will try to do that in my next PR. I know SQLite doesn't support the latest version of Electron, but I think if we stop using remote, keep all the DB calls in the browser, and use IPC events only for the renderer integration it probably should work (example https://github.com/sqlectron/sqlectron-gui/blob/656d868e1ed64829ea0fe5b6e01a38167d3eae37/src/browser/preload.ts). 

Anyway, that is a subject for the next PR, for now, this initial support should help us start improving the test coverage. We also will need unit tests, but I think that is gonna be simpler to do once we are not using the remote in the renderer anymore, so it is completely decoupled of the electron API.


MasterOdin:
Do you know the minimal version of Electron we would need to upgrade to support what you need?

maxcnunes:
By this report https://github.com/microsoft/playwright/issues/1042 it is broken on Chromium 80 and it should work on Chromium 82. Our current Electron version 8 is on Chromium 80 and the next version 9 is on Chromium 83. So, going to Electron 9 should work, although it would be awesome if we could go all over to the latest version.

MasterOdin:
Alright. I think moving to Electron 9 shouldn't be too much of an issue, but going to the latest might given that sqlite3 library is only tested up to Electron 8 and I believe one of the versions between 9 and latest changed how native extensions work.

maxcnunes:
Sure, that is why I mentioned that approach I would like to test. In case that doesn't work then I will try version 9.

maxcnunes:
@MasterOdin I got the e2e testing working for Sqlite basic operations:
* Connect to Sqlite DB
* Assert tables list
* Assert default select query
* Assert rows response (this one for some reason isn't working on the CI, I left it disabled for now on the CI)

I used another API method of playwright to avoid the issue I described before, so it wasn't required to upgrade electron because of that.

test/e2e/test.sqlite.js
```diff
@@ -0,0 +1,106 @@
...
+
+    // Assert rows
+    const rows = await this.mainWindow.$$('.ReactVirtualized__Grid__cell');
+    // NOTE: Keeping it disabled on CI for now. For some reason on running this
+    // assertion on CI it doesn't return any rows.
+    if (process.env.CI !== 'true') {
+      expect(rows).to.have.lengthOf(11); // rows + info header
+      for (let i = 0; i < rows.length; i++) {
+        await expect(await rows[i].innerText()).to.be.equal(i === 0 ? 'info' : `Ipsum ${i - 1}`);
+      }
```
MasterOdin, line 103:
```suggestion
      expect(await rows[0].innerText()).to.be.equal('info');
      for (let i = 1; i < rows.length; i++) {
        await expect(await rows[i].innerText()).to.be.equal(`Ipsum ${i - 1}`);
      }
```

test/e2e/test.sqlite.js
```diff
@@ -0,0 +1,106 @@
...
+  }
+
+  // Set DB path, it is required because sqlectron only suports absolute path
+  const configSample = fs.readFileSync(CONFIG_SAMPLE_PATH, { encoding: 'utf8' });
+  const config = JSON.parse(configSample);
+  config.servers[0].database = DB_PATH;
+  fs.writeFileSync(CONFIG_PATH, JSON.stringify(config, null, 2));
+
+  // Create database
+  var db = new sqlite3.Database(DB_PATH);
```
MasterOdin, line 26:
```suggestion
  const db = new sqlite3.Database(DB_PATH);
```

test/e2e/test.sqlite.js
```diff
@@ -0,0 +1,106 @@
...
+
+  db.serialize(function () {
+    db.run('CREATE TABLE document (info TEXT)');
+
+    var stmt = db.prepare('INSERT INTO document VALUES (?)');
```
MasterOdin, line 31:
```suggestion
    const stmt = db.prepare('INSERT INTO document VALUES (?)');
```

test/e2e/test.sqlite.js
```diff
@@ -0,0 +1,106 @@
...
+    for (var i = 0; i < 10; i++) {
```
MasterOdin, line 32:
```suggestion
    for (let i = 0; i < 10; i++) {
```

src/renderer/components/database-item.jsx
```diff
@@ -169,7 +169,10 @@ export default class DatabaseItem extends Component {
 
     return (
       <div>
-        <span style={style} className="item" onContextMenu={this.onContextMenu}>
+        <span
+          style={style}
+          className={`item item${dbObjectType}`}
```
MasterOdin, line 174: I like having a divider here between the class and the metatype, similar to how other libraries do it, eg. `item__${dbObjectType}` or `item-${dbObjectType}`.

MasterOdin (approved):
Some small changes, otherwise I think it looks good

MasterOdin:
With regards to electron and sqlite3, I would somewhat hold off on doing any of that. I am potentially leaning towards removing node-sqlite3 for better-sqlite3 depending on getting duplicate column results to work. I know that node-sqlite3 does not, and given the level of maintainer support there, I'm not confident it would anytime soon, so seriously considering better-sqlite3, in which case it might make easier the electron upgrade stuff.

maxcnunes:
@MasterOdin Alright, I can hold off that upgrade. Although, I probably will go ahead and prepare the code to support recent Electron version which is not using remote in the renderer and use direct IPC calls instead ([example](https://github.com/sqlectron/sqlectron-gui/blob/656d868e1ed64829ea0fe5b6e01a38167d3eae37/src/browser/preload.ts)).

PR #587, opened 2021-03-03
Fix linux build

maxcnunes:
Fix the linux build: https://github.com/sqlectron/sqlectron-gui/runs/2018631610

```
Run sudo apt-get install libgnome-keyring-dev icnsutils graphicsmagick rpm bsdtar
Reading package lists...
Building dependency tree...
Reading state information...
Package bsdtar is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
is only available from another source
However the following packages replace it:
  libarchive-tools
````

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -12,15 +12,16 @@ jobs:
     runs-on: ${{ matrix.os }}
 
     env:
-      node-version: 14
+      node-version: 10
 
     strategy:
       matrix:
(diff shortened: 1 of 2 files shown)


maxcnunes:
Now it is failing because of `libgnome-keyring-dev` https://github.com/electron/electron/issues/27704

MasterOdin:
I think it'd be easier to just specify `ubuntu-18.04` in the build matrix for now?

MasterOdin (approved):

PR #588, opened 2021-03-03
Add Typescript support

maxcnunes:
* Added Typescript support
* Converted all `src/browser` code to Typescript
* Converted `src/renderer/entry.tsx` and `src/renderer/store/configure.ts` to Typescript to ensure renderer code was supporting Typescript.
* Added a `src/common/types` to share types between renderer and browser code

I tested the app still working fine in dev mode and also with the dist app for Mac.


package.json
```diff
@@ -158,11 +173,14 @@
     "react-virtualized": "^9.22.3",
     "redux": "^3.5.0",
     "redux-logger": "^3.0.6",
-    "redux-thunk": "^2.3.0",
+    "redux-thunk": "^2.2.0",
```
maxcnunes, line 176: I had to downgrade a minor version because of a typing issue https://github.com/reduxjs/redux-thunk/issues/205

src/browser/config.ts
```diff
@@ -5,23 +5,25 @@
  * without passing through a transpiler, this file must use ES5.
  */
 
-const fs = require('fs');
-const path = require('path');
-const defaultsDeep = require('lodash.defaultsdeep');
-const sqlectron = require('sqlectron-core');
+import * as fs from 'fs';
```
MasterOdin, line 8: for this (and other imports from node stdlib), given that we have `esModuleInterop ` enabled, we should be able to just do `import fs from 'fs';`? Or is this required to work with babel?

MasterOdin (approved):
Looks good, couple small comments.

src/browser/config.ts
maxcnunes: I think it was failing even with that enabled, but let me try it once again

PR #592, opened 2021-03-04
Fix lint issues for src/browser

maxcnunes:
I noticed that the linter wasn't detecting many typing issues for `src/browser` code because the `@typescript-eslint/explicit-module-boundary-types` rule was turned off. 

PS: That same rule still turned off for `src/renderer` because turning it on now probably would require converting all files to typescript, or at least have a bunch of lint issues to be fixed right now. Leaving it for when/if we decide to convert those files to typescript (new files we should use typescript, but I don't mind keeping old ones in JS for now).

### DIFF
--- a/src/browser/core/servers.ts
+++ b/src/browser/core/servers.ts
@@ -2,13 +2,14 @@ import { v4 as uuidv4 } from 'uuid';
 import { validate, validateUniqueId } from './validators/server';
 import * as config from './config';
 import * as crypto from './crypto';
+import { Server, EncryptedPassword } from '../../common/types/server';
 
-export async function getAll() {
+export async function getAll(): Promise<Array<Server>> {
   const { servers } = await config.get();
   return servers;
 }
 
-export async function add(server, cryptoSecret) {
+export async function add(server: Server, cryptoSecret: string): Promise<Server> {
   let srv = { ...server };
   await validate(srv);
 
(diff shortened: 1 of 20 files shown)


src/browser/core/servers.ts
```diff
@@ -102,22 +103,30 @@ function encryptSecrects(server, cryptoSecret, oldServer?) {
...
 
-  if (server.password && typeof server.password === 'string') {
-    updatedServer.password = crypto.unsafeDecrypt(server.password, cryptoSecret);
-  } else if (server.password) {
-    updatedServer.password = crypto.decrypt(server.password, cryptoSecret);
+  if (server.password) {
+    if (typeof server.password === 'string') {
+      updatedServer.password = crypto.unsafeDecrypt(server.password, cryptoSecret);
+    }
+    {
```
MasterOdin, line 116:
```suggestion
    } else {
```

src/browser/menu/darwin.ts
```diff
@@ -1,18 +1,31 @@
...
 
-function sendMessage(win, message) {
+function sendMessage(win: BrowserWindow, message: string) {
   if (win) {
     win.webContents.send(message);
   }
 }
 
-export function buildTemplate(app, buildNewWindow, appConfig) {
+type BuildWindow = (app: App) => void; // eslint-disable-line no-unused-vars
```
MasterOdin, line 11: Move into `../../common`? Look's like we define this type a handful of times.

MasterOdin (approved):

src/browser/core/servers.ts
maxcnunes: wow, good catch

PR #593, opened 2021-03-05
Isolate the renderer process

maxcnunes:
In order to follow electron best security practices we should not use remote anymore in the renderer process. On its place we will use direct IPC calls specifically for what we need.
Sorry, it turned out to be a bit long PR, but it is doing a single thing: moved any electron/remote api usage from renderer process to the browser process and it is using ipc calls for that, but even the ipc calls aren't done directly through `ipcRenderer` module, instead every `ipcRenderer` is encapsulated by a Facade making it available to the frontend as a module.

* `src/renderer/api.ts`: is the API for every backend/browser integration
* `src/browser/preload.ts`: is where that API integrates with the backend/browser by ipcRenderer events
* `src/browser/ipcMain.ts`: is where those ipcRenderer events are received by the backend and proxied to the right module
* `src/common/types/api.ts`: has the interface `SqlectronAPI` for that integration, both renderer and process implement that to ensure the contract is properly followed and avoid any wrong implementation by them


src/browser/core/db.ts
```diff
@@ -0,0 +1,263 @@
...
+  setSelectLimit(limit: number): void {
+    return setSelectLimit(limit);
+  }
+}
+
+// Keep connections by BrowserWindow instances
+const connections: {
+  [windowId: string]: SqlectronDB;
+} = {};
+
```
maxcnunes, line 246: Because now the db connection lives in the browser process, it means when we there are multiple application windows open they share the same browser process, and to allow each window to have its own db connection, they are stored by windowId in this object.

src/browser/window.ts
```diff
@@ -39,7 +41,7 @@ export function buildNewWindow(app: App): void {
   // and load the index.html of the app.
   let entryBasePath = 'file://' + resolve(__dirname, '..');
   if (devMode) {
-    entryBasePath = 'http://localhost:8080';
+    entryBasePath = 'https://localhost:8080';
```
MasterOdin, line 44: Why do we need https here?

webpack.config.js
```diff
@@ -11,7 +11,7 @@ const isProd = process.env.NODE_ENV === 'production';
 const webpackConfig = {
   mode: isProd ? 'production' : 'development',
   devtool: 'eval-source-map',
-  target: 'electron-renderer',
+  target: 'web',
```
MasterOdin, line 14: Why the change away for electron as the target?

maxcnunes: The frontend code cannot access the electron module anymore, which turned the frontend more into a web app.

src/browser/window.ts
maxcnunes: Just to avoid warning security errors in the chrome console during development

MasterOdin (approved):
Small comments, otherwise looks good, good job on this :+1:

src/browser/window.ts
maxcnunes: @MasterOdin I just removed this change from this PR

src/browser/window.ts
```diff
@@ -29,8 +29,10 @@ export function buildNewWindow(app: App): void {
     minWidth: 512,
     minHeight: 350,
     webPreferences: {
+      nodeIntegration: false, // is default value after Electron v5
+      contextIsolation: true, // protect against prototype pollution
+      enableRemoteModule: false, // turn off remote
```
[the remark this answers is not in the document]
maxcnunes: Alright. Since the upgrading has been done in that other PR, I am gonna remove these properties there.

src/renderer/actions/config.ts
```diff
@@ -36,7 +32,8 @@ export function saveConfig(configData: Config): ThunkResult<void> {
     dispatch({ type: SAVE_CONFIG_REQUEST });
     try {
       await sqlectron.config.saveSettings(configData);
-      await sqlectron.setSelectLimit();
+      // TODO: Check, should it really be setting the limit wihtout a value?
+      sqlectron.db.setSelectLimit(1000);
```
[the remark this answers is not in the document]
maxcnunes: Done

src/browser/window.ts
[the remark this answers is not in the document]
maxcnunes: done

PR #595, opened 2021-03-06
Move actions and reducers to typescript

maxcnunes:
Moved actions and reducers to typescript. The best part is that I already caught a few issues that we were not aware.


src/renderer/actions/connections.ts
```diff
@@ -126,13 +135,13 @@ export function connect(id, databaseName, reconnecting = false, sshPassphrase) {
       }
       const currentConn = getState().connections;
       if (!currentConn.databases.length) {
-        this.disconnect();
+        dispatch(disconnect());
```
maxcnunes, line 138: Found 2 issues here:
1. It was calling this.disconnect, which does not exist, instead of disconnect
2. It was not dispatching the event from the disconnect execution

src/renderer/actions/queries.ts
```diff
@@ -102,18 +110,19 @@ function shouldUpdateQuery(query, selectedQuery, state) {
...
-export function appendQuery(query) {
+export function appendQuery(query: string): ThunkResult<void> {
   return (dispatch, getState) => {
-    const currentQuery = getCurrentQuery(getState()).query;
+    const queryState = getCurrentQuery(getState());
+    const currentQuery = queryState.query;
     const newLine = !currentQuery ? '' : '\n';
     const appendedQuery = `${currentQuery}${newLine}${query}`;
-    if (!currentQuery.isExecuting) {
+    if (!queryState.isExecuting) {
```
maxcnunes, line 119: Issue found: It was accessing `isExecuting` from the query text, which does not existing since it is a string, instead of the query state.

src/renderer/actions/queries.ts
```diff
@@ -274,9 +294,9 @@ export function cancelQuery(queryId) {
   };
 }
 
-function stringifyResultToCSV(origRows, delimiter) {
+function stringifyResultToCSV(origRows: [], delimiter: string): Promise<string> {
   if (!origRows.length) {
-    return '';
+    return Promise.resolve('');
```
maxcnunes, line 299: Issue found: Wasn't returning a promise when there are no rows while the caller expect it to always get a promise back.

src/renderer/store/configure.ts
```diff
@@ -46,10 +46,8 @@ if (isLogConsoleEnabled || isLogFileEnabled) {
   middlewares.push(createReduxLogger(loggerConfig));
 }
 
-const createStoreWithMiddleware = applyMiddleware(...middlewares)(createStore);
-
 export default function configureStore(initialState?) {
-  const store = createStoreWithMiddleware(rootReducer, initialState);
+  const store = createStore(rootReducer, initialState, applyMiddleware(...middlewares));
```
maxcnunes, line 50: No need to do what we were doing before, `createStore` accepts a third argument for middlewares.

PR #604, opened 2021-03-18
do not render table with if no fields or rows

### LINKED ISSUE #573: Blank Screen Make App Unusable
Some querys when performed causes the app to crash, also running CTRL+R or CTRL+ENTER have the same problem sometimes.
I am using Linux POP!_OS 20.04 LTS, but I have the same problem on a Windows 10 machine too.
image
image

MasterOdin:
Fixes #573

This fixes the crash where on running a query that's not recognized, but also returns an empty resultset, the app would crash. Now, we'll just print a nice generic "Query success" type message in-lieu of anything better.

### DIFF
--- a/src/renderer/components/query-result.jsx
+++ b/src/renderer/components/query-result.jsx
@@ -75,10 +75,22 @@ export default class QueryResult extends Component {
       const title = fields[0].name;
       return (
         <Message
-          key={queryIndex}
+          key={`explain-${queryIndex}`}
           preformatted
           title={title}
           message={rows.map((row) => row[title]).join('\n')}
+        />
+      );
+    }
+
+    // Not sure what type of query they ran, but cannot render table, print
+    // generic message.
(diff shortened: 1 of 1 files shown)


src/renderer/components/query-result.jsx
```diff
@@ -75,14 +75,26 @@ export default class QueryResult extends Component {
...
+    // generic message.
+    if (fields.length === 0) {
+      return (
+        <Message
+          key={`genericResult-${queryIndex}`}
+          message={`Query executed successfully.`}
+          type="success"
+        />
+      );
+    }
```
MasterOdin, line 96: I put this `if` statement here (as opposed to high-up as possible) to act as a final catch-all before rendering the table, and that we may add to the the possible conditions in which we enter this if statement, and do not want to worry about colliding with `explain` or anything else we might put above it,

maxcnunes (approved):
Good catch

PR #606, opened 2021-03-20
Fixed the Issue in darkmode when list connection is on then text color is black.

ma3ve:
Fixed the Issue in darkmode when list connection is on then text color of list is dark

What i Did: I get the enabledDarkTheme from props and added inline
styling based on that

### DIFF
--- a/src/renderer/components/server-list-item.jsx
+++ b/src/renderer/components/server-list-item.jsx
@@ -12,7 +12,7 @@ const ICONS = sqlectron.db.CLIENTS.reduce((clients, dbClient) => {
   return clients;
 }, {});
 
-const ServerListItem = ({ server, onConnectClick, onEditClick }) => (
+const ServerListItem = ({ server, onConnectClick, onEditClick, enabledDarkTheme }) => (
   <div className="item">
     <div className="middle aligned content">
       <div className="left floated" style={{ padding: '1em' }}>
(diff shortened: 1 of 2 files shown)


maxcnunes:
@ma3ve thanks for the contribution! Could you please just added a screenshot of before and after your changes please?

ma3ve:
Before
Before

After
Before

MasterOdin (changes requested):
Thanks for submitting this patch and interest in sqlectron!

I'm not a fan of this PR on two reasons.
* For style stuff, I would much rather see us modify the `*.css` and `*.scss` files first if possible since then we have a nice centralized place for these element definitions, and tweaking them is easier (as we do not need to muck about with react, context, etc.). For the context of this particular issue, we can hit an equivalent solution by amending https://github.com/sqlectron/sqlectron-gui/blob/main/src/renderer/components/server-list.scss and adding:

```scss
body.dark-theme #server-list {
  .header {
    color: #fff;
  }

  .meta {
    color: #fff;
  }
}
```
We then do not need to modify any of the `*.jsx` files.

* I'm not a fan of around theme around as a direct prop. The theme applies across a component tree, and to all its children, and if we were to go the route of conditional styling based on theme, we might have a lot more instances of where we would want to use the theme. This can create some amount of headache to us as we would then (using this example) have to then deal with long chains of passing theme around components to do this kind of direct styling. Better would probably be to use [React Context](https://reactjs.org/docs/context.html) where we want to share a prop across the tree. This would allow us to set it at the top-level `<App />` and then grab it where we need it. This is more a technical note, and not something I'm asking we do here (I would much rather just see the `server-list.scss` change).

MasterOdin:
Thanks again for submitting this PR as it did the brunt of the work on fixing this issue.

Having said that, I've gone ahead and implemented my suggestion above in #06a0e0f, so closing this.

Issue #550, opened 2020-11-12
Upgrade Electron 10


maxcnunes:
This will help us keep up to date with all the improvements done to Electron.

MasterOdin:
Will need to move to the newest version of the sqlite3 library (5.0.1), which breaks compatibility for node < 10. Additionally, given some of the bug reports there, it's not totally clear if moving to 10 will work, or if 9 is as high as we could go and make sure that `allowRendererReuse` is false.

Issue #471, opened 2019-06-07
Cannot parse privateKey: Unsupported key format


robertgarrigos:
I cannot connect to a server thru ssh tunnel and and openssh key. Got this error: 
```
Connection Error
Cannot parse privateKey: Unsupported key format
```
It's the same key I'm using to connect thru ssh to my server.

I'm on an Ubuntu 18.04.2 LTS and latest sqlectron as of today. Any ideas?

Thanks.

btekbtek:
I had this issue today, You need to use id_rsa.pem
if You don't have, use any converter.

btekbtek:
Ok Finally I get this to work.

What I done on my Ubuntu:
1. I use 'wine puttygen.exe'
2. convert my id_rsa to id_rsa.ppk
3. this not work so I convert id_rsa.ppk to id_rsa.pem

and its connect.

natethor:
I'm also getting this issue.  I created my key normally on OSX terminal.  I'd rather not have to do the work-around above using wine and puttygen.  Any idea if there's plan for a fix?

MasterOdin:
I've added this to the backlog for the next version.

armarti:
> Ok Finally I get this to work.
> 
> What I done on my Ubuntu:
> 
> 1. I use 'wine puttygen.exe'
> 2. convert my id_rsa to id_rsa.ppk
> 3. this not work so I convert id_rsa.ppk to id_rsa.pem
> 
> and its connect.

For me the command below _nearly_ worked. Perhaps it can help someone get to the complete solution quicker.

Using puttygen-0.74 on macOS (`brew install putty`):

```
puttygen ./ssh-keygen-created.private.key -O private-openssh -o ./plain-rsa.private.key
```

Though it only connects successfully using this `./plain-rsa.private.key` when the key is decrypted. And even then it only connects successfully when clicking the **Test** button while creating the connection. When I go to to actually connect to see the tables and whatnot I get the same `Cannot parse privateKey: Unsupported key format`. Same results with rsa2048, rsa4096, and ed25519 keys.

This `puttygen` makes a distinction between **openssh** and **openssh-new** formats.

```
$ puttygen --help
...
  -O    specify output type:
           ...
           private-openssh     export OpenSSH private key
           private-openssh-new export OpenSSH private key (force new format)
...
```

The "openssh" format starts with:

```
-----BEGIN RSA PRIVATE KEY-----
```

The "openssh-new" format starts with:

```
[...]

PR #607, opened 2021-03-23
Upgrade electron to v12

maxcnunes:
Closes https://github.com/sqlectron/sqlectron-gui/issues/550

Now that we are not using remote anymore for sqlectron-db-core calls, and instead we are using IPC calls and running that in the browser process, which is a node process, there is no need anymore to build native dependencies to Electron anymore. It allow us to upgrade to latest version of Electron, ~which is actually required to fix https://github.com/sqlectron/sqlectron-gui/issues/471 because a new version of ssh2 that supports new auth types depends on Node12.~ ([just noticed Electron v8 is already on Node12](https://www.electronjs.org/releases/stable?version=8&page=5#release-notes-for-v800)).

I tested it running from source and a compiled version for Mac, and both worked fine with MySQL and Sqlite databases.

PS: It is pointing to ipc branch because that has not been merged yet, I am going to point it to master once that has been merged.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6081,14 +6081,22 @@
       }
     },
     "electron": {
-      "version": "8.5.5",
-      "resolved": "https://registry.npmjs.org/electron/-/electron-8.5.5.tgz",
-      "integrity": "sha512-e355H+tRDial0m+X2v+l+0SnaATAPw4sNjv9qmdk/6MJz/glteVJwVJEnxTjPfEELIJSChrBWDBVpjdDvoBF4Q==",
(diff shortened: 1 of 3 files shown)


MasterOdin:
Yeah, don't think we're affected by any breaking changes? Would be good to confirm stuff like menus and opening/saving queries also works under electron 12 along with the database connector stuff.

maxcnunes:
@MasterOdin I have tested everything I could think of, including those you mentioned and couldn't find any issues so far.

MasterOdin (approved):

PR #624, opened 2021-03-29
Add ssh agent option

### LINKED ISSUE #471: Cannot parse privateKey: Unsupported key format

maxcnunes:
Will close #471 

Working in progress, still need to add tests


docker-compose.yml
```diff
@@ -1,12 +1,18 @@
...
-  # because is not possible packing for OSX from Linux.
-  # Use unsafe-perm to force npm respect the tmp path (http://git.io/vB2oR)
-  command: npm run --unsafe-perm dist:winlinux
-  environment:
-    # Keep the tmp folder inside the project.
-    # This way we dont lose some cache files between distribution tasks
-    TMPDIR: /usr/src/app/.tmp
-  volumes:
-    - .:/usr/src/app
+version: '3'
```
MasterOdin, line 1: Are you still using this file for anything? Should it be removed?

MasterOdin (approved):
Changes make sense. Are you planning on adding another test for the new ssh agent setting, since it looks like the E2E tested added here is for SSH through a private key file.

docker-compose.yml
maxcnunes: It was used for builds when we just had Ubuntu OS in the CI. But, it maybe makes sense keeping it around just in case we need to test a build locally for other OS than our own host OS.

maxcnunes:
The intention was adding the test for basic private key only because of the fixing for private keys which comes with the db-core upgrade. But it should be simple to add a test for ssh agent as well. I will do it later today after working hours.

Issue #612, opened 2021-03-25
Support import connection string


maxcnunes:
It simplifies creating a new connection when we already have the connection URI-Like string.

PR #625, opened 2021-03-30
Add connection URI support

maxcnunes:
Will close https://github.com/sqlectron/sqlectron-gui/issues/612

This PR adds a URI field at the bottom of the database info. That field works in 2 ways:
1. If the user changes that field directly, it will update the database info details.
2. If the user changes the database fields, the URI field will be updated based on those changes.

Also, this URI field respects the password hidden button. So, it will sanitize the password in the URI field based on that button state.

Screen Shot 2021-05-01 at 10 34 27


MasterOdin (approved):
Seems to work as expected. Only nitpick would be that on pasting a URI, the password in the URI is shown to the user while hidden in the password field. We may want to either hide the field within the URI or set `showPlainPassword` to true to keep them in-sync?

maxcnunes:
@MasterOdin good catch. I disabled the URI field when the password is hidden, and it will show a message to the user, in that case, to make it clear the password must be visible to change that field. Please, let me know if you can think of a better message than what I used there.
Screen Shot 2021-05-01 at 18 01 50
Screen Shot 2021-05-01 at 18 01 39

MasterOdin (approved):
I'm good with this. Having it be `disabled` keeps it still copyable if so desired.

vitaly-t:
Hello!

I am the author of `connection-string` library.

See the following PR with some improvements: [PR-644](https://github.com/sqlectron/sqlectron-gui/pull/644).

Issue #617, opened 2021-03-26
Electron Security Warning


maxcnunes:
Follow Electron best security practices (https://www.electronjs.org/docs/tutorial/security) and take care of this warning message:

<img width="490" alt="Screen Shot 2021-03-26 at 18 52 55" src="https://user-images.githubusercontent.com/680356/112696421-8ab62a00-8e64-11eb-9f0f-d6363b1ada7a.png">

Issue #533, opened 2020-10-22
Update to jointjs 3.x


MasterOdin:
Updating to jointjs 3.x as-is leads to a break in functionality. I suspect it has to do with this line from their changelog (though I'm not sure):

> * remove deprecated `PortsModelInterface` and `PortsViewInterface` from `joint.shapes.basic` [breaking change]

In any case, when upgrading to 3.x, it causes the diagram to generate blank looking boxes.

maxcnunes:
> In any case, when upgrading to 3.x, it causes the diagram to generate blank looking boxes.

Just FYI, It was caused by the namespacing change, which needs to be defined on v3, check out the PR #664 for more details.

PR #638, opened 2021-04-17
Allow unsafe-eval CSP for jointjs

MasterOdin:
"Generate Database Diagram" was broken on the inclusion of a CSP (#623) as it requires the `unsafe-eval` policy. This policy is not recommended for security reasons, but is necessary until we can upgrade to jointjs (#533). Given that we do not load external webpages, the risk here is somewhat minimal until we can resolve it.

This does re-introduce the warning in #617, but the alternatives are fix the upgrade to JointJS 3.x (which is beyond my time constraints at the moment) or just disable the diagram rendering altogether.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -105,7 +105,7 @@ const webpackConfig = {
       {
         'base-uri': "'self'",
         'object-src': "'none'",
-        'script-src': ["'self'"],
+        'script-src': ["'unsafe-eval'", "'self'"],
         'style-src': ["'unsafe-inline'", "'self'"],
       },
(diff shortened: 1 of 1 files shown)


maxcnunes:
As we discussed on Slack, upgrading JointJS to the latest version would require too much work and it is in none of our scopes to work on that. Also, we are not sure if anyone is even using that feature. So, it seems it would be better to disable or remove the diagram feature for now and keep the app in a good security state. Once we have time to add the diagram back again, we could upgrade JointJS or even use a different framework.

Issue #630, opened 2021-04-04
Errors not visible in the server management window


maxcnunes:
Looks like when we moved from remote to IPC calls the errors in the server management window are not visible anymore.

maxcnunes:
I found out this issue occurs because we use a custom error object for server validation failures, and since we have changed the app to use IPC calls only, when the error is thrown from the browser to the renderer process, the error object is serialized and only the error built-in props are copied over that serialization https://github.com/electron/electron/issues/24427

PR #642, opened 2021-05-01
Fix server validation not showing errors in the UI

maxcnunes:
Will fix https://github.com/sqlectron/sqlectron-gui/issues/630

Because of the issue described on https://github.com/sqlectron/sqlectron-gui/issues/630, we cannot pass custom properties through the error object anymore. As far as I know, only the server validations suffer from this problem.

### DIFF
--- a/src/browser/core/servers.ts
+++ b/src/browser/core/servers.ts
@@ -2,16 +2,20 @@ import { v4 as uuidv4 } from 'uuid';
 import { validate, validateUniqueId } from './validators/server';
 import * as config from './config';
 import * as crypto from './crypto';
-import { Server, EncryptedPassword } from '../../common/types/server';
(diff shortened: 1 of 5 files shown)


src/browser/core/servers.ts
```diff
@@ -41,10 +49,10 @@ export async function update(server: Server, cryptoSecret: string): Promise<Serv
 
   await config.save(data);
 
-  return server;
+  return { data: srv };
```
maxcnunes, line 52: While working on this ticket, I noticed a bug in this code, we were returning the wrong object. It should be returning the saved record, which has the encrypted password, just like we do for the `add` function, but it was returning the incoming data with the decrypted password.

MasterOdin (approved):

PR #644, opened 2021-05-02
Improve ConnectionString usage

vitaly-t:
As a follow-up [to this](https://github.com/sqlectron/sqlectron-gui/pull/625), from the author of [connection-string] ;)

* Safer and simpler way to access the first host details. See [Virtual Properties](https://github.com/vitaly-t/connection-string#virtual-properties).

* The initial code tried to re-implement `passwordHash` option that [connection-string] supports. It also changes the default `#` symbol to `*` instead, to keep it consistent with how original code worked.

[connection-string]:https://github.com/vitaly-t/connection-string


maxcnunes (approved):
That looks great. Thanks for the simplification!

Issue #643, opened 2021-05-02
"enabledLiveAutoComplete": false has no effect


betafcc:
I'm using 1.36.0, macos, MySql 8 database (but I noticed on others as well).

I'm trying to disable the live completion but it just won't go
I tried both using the settings on the gui and editing sqlelectron.json, neither `"enabledLiveAutoComplete": false` nor ` "enabledAutoComplete": false` makes any effect, when I go to my database and start typing in the query editor, immediately it starts appearing.

It is quite annoying specially because I can't unset the 'enter commit completion', making it quite unpleasant to type anything before a newline.

Here is the full config file:

```json
{
  "servers": [
    {
      "id": "cb508927-9f47-41b7-b457-150425a2045a",
      "name": "mysql",
      "client": "mysql",
      "ssl": false,
      "host": "localhost",
      "port": 3306,
      "socketPath": null,
      "user": "root",
      "password": {
        "ivText": "6lv2/JcFt9rMLiANclLr0A==",
        "encryptedText": "GIYgFDGkmI0ZuYBBdK8Prg=="
      },
      "schema": null,
      "encrypted": true
    }
  ],
  "zoomFactor": 1,
  "limitQueryDefaultSelectTop": 100,
  "enabledDarkTheme": false,
  "disabledOpenAnimation": true,
  "csvDelimiter": ",",
  "connectionsAsList": false,
  "customFont": "Lato",
  "enabledAutoComplete": false,
  "enabledLiveAutoComplete": false
}
```

And the docker-compose file I'm using to run the db (although I doubt it makes any difference):

```
version: '3.3'
services:
  db:
    image: mysql:8
    command:
      - '--local-infile=1'
      - '--secure-file-priv='
    restart: always
    security_opt:
      - seccomp:unconfined
    environment:
      MYSQL_DATABASE: 'db'
      MYSQL_USER: 'user'
      MYSQL_PASSWORD: 'password'
      MYSQL_ROOT_PASSWORD: 'password'
    ports:
      - '3306:3306'
    expose:
      - '3306'
    volumes:
      - ./my-db:/var/lib/mysql
      - ./mysql-files:/var/lib/mysql-files
volumes:
  my-db:
  data:
```

Thanks for the otherwise awesome project!

maxcnunes:
Thanks for reporting it @betafcc . I just confirmed the app isn't respecting the configuration file values for those fields.

betafcc:
[...]

Issue #319, opened 2017-05-18
Impossible to disable autocompete


ederavilaprado:
Hello there.

I'm unable to turn-off the "autocomplete" functionality.
I changed my config settings following the documentation, but it did not work in the application.
Since our database is very large, this feature is leaving the application very slow and unfortunately, impossible to be used.

My scenario:
- OS: MacOs Sierra 10.12.4 (16E195)
- Sqlectron: Version 1.20.2 (1.20.2.439)
- Database: Postgres

This is my config file `~/.sqlectron.json`
```json
{
  "enabledAutoComplete": false,
  "enabledLiveAutoComplete": false,
  "servers": [
    {
      "id": "5070bc14-62f8-4be6-a973-8a6ecaa9cf7d",
      "name": "my_connection",
      "client": "postgresql",
      "ssl": false,
      "host": "127.0.0.1",
      "port": 5432,
      "socketPath": null,
      "user": "userx",
      "password": "123",
      "database": "dby",
      "schema": null
    }
  ]
}
```

Thanks in advance for this great app

maxcnunes:
Does the autocomplete still appearing for you? It shouldn't.

Do you need access all schemas or tables available? You could filter them with `filter` configuration (also described [here](https://github.com/sqlectron/sqlectron-gui/blob/master/docs/app/configuration-file.md)).

ederavilaprado:
Hi @maxcnunes.

I did some tests here and your tip worked perfectly for me, but doing in a hack way.
I tested several different filters combinations, but the input lag persists (a very high lag).

What i did was to disable the autocomplete filtering this way.
```json
"filter": {
  "database": {
    "only": [
      ""
    ]
  }
}
```

But anyways, the problem with the flags `enabledAutoComplete`and `enabledLiveAutoComplete` persists.

I'm not sure if i can help trying to fix the code and opening a PR, but i will try this weekend.

MasterOdin:
This should have been fixed quite a while ago as part of #645.

PR #645, opened 2021-05-03
Fix query configurations from file

maxcnunes:
Will fix https://github.com/sqlectron/sqlectron-gui/issues/643 and https://github.com/sqlectron/sqlectron-gui/issues/319.

It was accessing the configuration properties directly from `action.config`, but the configuration data is actually at `action.config.data`.

### DIFF
--- a/src/renderer/reducers/queries.ts
+++ b/src/renderer/reducers/queries.ts
@@ -1,9 +1,25 @@
-import { Action, Reducer } from 'redux';
+import { Action, Reducer, AnyAction } from 'redux';
+import { ConnectionAction } from './connections';
 import * as connTypes from '../actions/connections';
 import * as types from '../actions/queries';
 
+interface QueryPartialUpdate {
+  name?: string;
+  filename?: null | string;
+  isExecuting?: boolean;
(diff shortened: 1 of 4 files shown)


src/renderer/reducers/queries.ts
```diff
@@ -201,30 +223,40 @@ const queryReducer: Reducer<QueryState> = function (
...
-    return state;
-  }
+function addNewQuery(
+  state: QueryState,
+  queryAction: QueryAction | null,
+  connAction?: ConnectionAction,
+): QueryState {
+  let { enabledAutoComplete, enabledLiveAutoComplete, resultItemsPerPage } = INITIAL_STATE;
 
-  const configItemsPerPage = action.config && action.config.resultItemsPerPage;
```
maxcnunes, line 209: I couldn't find any reference for this field `resultItemsPerPage` in the Config object, so changed it to only rely on the state object for this field.

But, even though the query state object has a definition for this field, it is not really being used anywhere. We pass it down to some components, but it doesn't look like it is really being used. So, looks like we could drop off it field completely.

MasterOdin (approved):

PR #651, opened 2021-05-07
Upgrade react-tabs

maxcnunes:
There were no breaking changes for us.


MasterOdin (approved):
Looks like the breaking change between 2.x and 3.x was just in the minimum supported react version (16.3) which is below what sqlectron is using.

Issue #657, opened 2021-05-08
Move path module dependency to the browser process


maxcnunes:
As part of the migration to Webpack 5 it was required to add a polyfill for the `path` module because it is used in a few places in the renderer process. This task should consider moving the logic with that dependency to the browser process, and by that avoid any node polyfill for the renderer process.

Look for this ticket number in the code base, it is being referenced where those configurations were done.

PR #662, opened 2021-05-22
Remove path browserify moving logic to the browser process

maxcnunes:
Will close https://github.com/sqlectron/sqlectron-gui/issues/657


MasterOdin (changes requested):
Looks like there's at least two imports (path and fileHandler) that are never used and can be removed from `src/renderer/actions/queries.ts`, one of which is the unused `path` that we're trying to remove from usage in the frontend. While webpack does get rid of it during compilation, would probably be good to not have it show up in the code either.

maxcnunes:
Good catch. I would expect that kind of issue to be caught by eslint, but just realized `no-unused-vars` rule is turned off. I will leave it to another PR to check whether we can enable it back again.

Issue #618, opened 2021-03-26
NPM dependency vulnerabilities


maxcnunes:
On installing the app dependencies with npm it shows that there are vulnerabilities in some of our dependencies. We should try to handle them.

MasterOdin:
They're mostly all centered around jointjs. Upgrading it (#533) should resolve, but there's some work that is required there.

PR #664, opened 2021-05-24
Upgrade deps available to reduce vulnerabilities

maxcnunes:
Will close https://github.com/sqlectron/sqlectron-gui/issues/533 and https://github.com/sqlectron/sqlectron-gui/issues/618.

Upgrade a few dependencies to reduce the number of vulnerabilities reported by NPM audit.

The most important update is for JointJS which required allowing unsafe-eval in CSP rules (more details in https://github.com/sqlectron/sqlectron-gui/issues/663).

With this upgrading, now NPM audit has just a few vulnerabilities listed and it seems they are all from storybook deps, which is only a dev dependency:

```
found 53 vulnerabilities (49 moderate, 4 high) in 3826 scanned packages
```


src/renderer/components/database-diagram.jsx
```diff
@@ -20,7 +18,7 @@ export default class DatabaseDiagram extends Component {
   constructor(props) {
     super(props);
     this.state = {};
-    this.graph = new joint.dia.Graph();
+    this.graph = new joint.dia.Graph({}, { cellNamespace });
```
maxcnunes, line 21: It was required to use this namespace definition for custom elements on version 3 of JointJS https://stackoverflow.com/a/57381434/1050818

src/renderer/components/database-diagram.jsx
```diff
@@ -84,25 +83,26 @@ export default class DatabaseDiagram extends Component {
...
+            y: 20 + (index % 4) * 100,
+          },
+          size: {
+            width: 120,
+            height: (columnsByTable[table].length + 1.5) * 20,
+          },
+          name: table,
+        });
+
+        tableShapes.push(tableShape);
```
maxcnunes, line 98: Same code, just moved the table definition into a variable before pushing it into the array

src/renderer/components/database-diagram-shapes.js
```diff
@@ -0,0 +1,113 @@
+import * as joint from 'jointjs';
```
maxcnunes, line 1: Moved all custom shape definitions into the same file

webpack.config.js
```diff
@@ -110,8 +110,11 @@ const webpackConfig = {
       {
         'base-uri': "'self'",
         'object-src': "'none'",
-        'script-src': ["'self'"],
+        // NOTE: unsafe-eval is required because of JointJS
+        // https://github.com/sqlectron/sqlectron-gui/issues/663
+        'script-src': ["'self'", "'unsafe-eval'"],
```
maxcnunes, line 115: We talked about not doing this, but I thought about it again and it seemed worse losing the diagram feature just because of the CSP warning. I think it is better just ignore the warning for now, and work on changing the diagram implementation later to avoid this security issue.

package.json
```diff
@@ -180,7 +180,8 @@
     "node-sass": "^5.0.0",
     "playwright-core": "^1.9.1",
     "playwright-electron": "^0.5.0",
-    "postcss-loader": "^3.0.0",
+    "postcss": "^8.3.0",
```
maxcnunes, line 183: The breaking change of `postcss-loader` moved `postcss` to a peer dependency, that is why I added it here.

package.json
```diff
@@ -140,12 +140,12 @@
     "concurrently": "^2.2.0",
     "cross-env": "^7.0.2",
     "csp-html-webpack-plugin": "^5.1.0",
-    "css-loader": "^4.2.1",
+    "css-loader": "^5.2.5",
```
maxcnunes, line 143: No breaking changes for us.

package.json
```diff
@@ -131,7 +131,7 @@
     "@types/webpack-env": "^1.16.0",
     "@typescript-eslint/eslint-plugin": "^4.16.1",
     "@typescript-eslint/parser": "^4.16.1",
-    "autoprefixer": "^9.8.0",
+    "autoprefixer": "^10.2.5",
```
maxcnunes, line 134: No breaking changes for us.

.github/workflows/build.yml
```diff
@@ -12,7 +12,7 @@ jobs:
     runs-on: ${{ matrix.os }}
 
     env:
-      node-version: 10
+      node-version: 14
```
maxcnunes, line 15: Required to fix typescript build https://stackoverflow.com/questions/64725249/fs-promises-api-in-typescript-not-compiling-in-javascript-correctly

maxcnunes:
@MasterOdin I intend to release a new version once this change is merged

MasterOdin:
I'm not convinced it's worth keeping the diagram feature around. Testing out this PR, sqlectron generates the following diagram

Screen Shot 2021-05-25 at 1 04 50 AM

This is the same though that 1.31.0 also generated, so it would seem this has been broken for some time, making me wonder if anyone is actually using this feature.

For reference, DBBeaver generates the following diagram:

Screen Shot 2021-05-25 at 1 07 24 AM

maxcnunes:
I was testing on MySQL only, which seems to be generating the diagram properly:

Screen Shot 2021-05-25 at 08 31 15

I just tested it on Postgres and I get the same error as you. I guess this is an issue for adapters with schema support (PG, MSSQL). That means if anyone is really using this feature it is only on MySQL otherwise we would have got a report for it already. Since it is already like that, I would like to get it merged as it is now, and I can create a ticket specifically for the fixing to support the other adpters.

PR #682, opened 2021-09-15
Update build/icon.icns to conform to Big Sur style

amancevice:
Updates the app icon to conform to macOS Big Sur style:

Screen Shot 2021-09-15 at 3 42 57 PM


MasterOdin:
Thanks for this! The icon looks great!

The only concern I'd have is that it does boost the size of the icon quite a bit (especially compared to other icons), is there a way to reduce the size at all?

amancevice:
Great question @MasterOdin ! I'm not an icon expert — I used a tool called [image2icon](https://img2icnsapp.com) to generate the file. It _looks_ like maybe it just has too many resolutions packed into it? I'll see if I can extract a couple of them.

MasterOdin:
Thanks for doing this!

PR #687, opened 2022-01-20
Fixed bug where switching tab causes loss of focus on editor

synle:
### Problem 
Currently if you have multiple tabs, switching to other tab does not automatically focus on the ace editor (used for query). Though  the editor itself has a caret indicator saying it's focused on. So it is kind of confusing.

I checked the code and found that there is a `autoFocus={true}` property set on the react-ace, but it's not working properly when switching tab.

# Screenshots
image

Notice the caret at the end of the select statement
image


MasterOdin:
Thanks!

PR #710, opened 2022-07-16
fix(query-browser): remove the height prop from ResizableBox

ruddenchaux:
update react-resizable dependency

✅ Closes: 709


MasterOdin (approved):
Thanks for the fix!

PR #723, opened 2023-05-27
Fix search crash

Ormax563:
Added a function (Regex String) to remove some special characters from a string before passing it to the RegEx function.


MasterOdin (approved):
Thanks for your contribution. :+1:

PR #744, opened 2026-03-17
Replace env-paths with os-user-dirs

piroz:
### Motivation

\`env-paths\` has not been updated in 4+ years. \`os-user-dirs\` is an actively maintained
alternative that provides the same \`projectDirs()\` API, plus additional features:

- **CJS + ESM dual support** (env-paths v3 is ESM-only)
- **Zero dependencies** (same as env-paths)
- **Additional APIs**: user directories, XDG base directories, system search paths
- **TypeScript included**
- **Active maintenance**: regular releases in 2026

### Changes

- Replaced \`env-paths\` with \`os-user-dirs\` in \`package.json\`
- Updated import in \`src/browser/core/utils/index.ts\`
- Use \`projectDirs('Sqlectron').config\` instead of \`envPaths('Sqlectron', { suffix: '' }).config\`

\`projectDirs()\` defaults to no suffix, so the behavior is identical.

### Migration Reference

See: https://github.com/velocitylabo/os-user-dirs/blob/master/docs/migration-from-env-paths.md

### Testing

- [ ] All existing tests pass
- [ ] Verified on Linux/macOS/Windows

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -12,7 +12,7 @@ jobs:
     runs-on: ${{ matrix.os }}
 
     env:
-      node-version: 18
+      node-version: 20
 
     strategy:
       matrix:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -12,7 +12,7 @@ jobs:
     runs-on: ${{ matrix.os }}
 
     env:
-      node-version: 18
+      node-version: 20
 
     strategy:
       matrix:
--- a/package-lock.json
+++ b/package-lock.json
@@ -17731,6 +17731,21 @@
         "node": ">= 4"
       }
     },
+    "node_modules/@storybook/react/node_modules/type-fest": {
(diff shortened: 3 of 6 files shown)


MasterOdin:
Thank you for your PR. While we're still using a very old version of `env-paths`, it does still get the occasional update (it's on v4 as of 5 months ago). However, from what I could tell, the core of it hasn't changed really from v1 to v4 except for modifying where it saves files for Windows (going from `%APPLOCALDATA%` to `%APPDATA%`). So to update, we'd need to handle that, granted looking at your link, we'd need to handle that path move as well as for macOS, as we cannot just lose the `sqlectron.json` file that we save in the `.config` directory.

Having said that, given our extremely slim usage of `env-paths`, I'm also not sure I see the value add of moving to `os-user-dirs` at the moment. 

I'll add that I don't view EJS only as a negative for `env-paths`, and do agree with sindresorhus philosophy on looking to push everyone to using it.

